PR #75, opened 2021-06-21
Overview article for the stake pool section

Nazeim:


katomm (changes requested):
It feels like the PR is changing the whole structure without discussing it first:

- Prerequisites placeholder was replaced with no prerequisites.
- Security placeholder was replaced with nothing security-related.
- Any reason why the marketing placeholder was not replaced with the marketing article but a forum post? 

Thanks!

katomm (changes requested):
Perhaps the "prerequisites"-placeholder was not written clearly enough.

In terms of the basic requirements for running a pool, we should be talking about knowledge and experience. Linux and networking skills, terminal and cli stuff, the ability to implement best practices, etc, we also want to manage expectations that not everyone should run a pool unless these people are willing to learn and read a ton. This is not easy content and should be written by good SPOs. 

I recommend closing this PR now and leaving the content as it is with the placeholders. Thanks!

PR #127, opened 2021-07-13
Add crowdano

ejane24:


katomm:
Thank you for participating here. Unfortunately, this pull request needs to be closed because of:

* Project does not run on Cardano mainnet
* Project has no running product

PR #143, opened 2021-07-15
fix markdown warnings in styleguide

tweakch:
- I fixed all markdown warnings (where it was possible) in the style guide and
- added a section for editor extensions one could use to be more productive 
- added `php` linting to the Client because there is a code-snippet in the styleguide  
- and typos


katomm (approved):
Thank you very much for your contributions. 🙌🏻

Issue #152, opened 2021-07-16
Use consistent paths in code samples (accross all guides)


tweakch:
It would make sense to agree on a consistent environment across guides

From what I've seen so far, it's *pretty consistent* but not everywhere... I find that I have to switch directory structure between different guides and my files started ending up everywhere... 

I would suggest the following setup

```text
/home/user/cardano
|_ /keys # from cardano-cli guides
|_ /wallets # from cardano-wallet guides
|_ /db
|_ /integrate-cardano-samples
  |_ /id-of-guide-1
    |_ draft.tx
    |_ draft.signed
  |_ /id-of-guide-2
  |_ ...
|_ /build-tx-metadata-samples
  |_ /id-of-guide-1
  |_ /id-of-guide-2
  |_ ...
``` 

This setup would allow all guides to first ask the user to check that he is in the correct directory (and let him `cd` into it)

```
$ pwd
/home/user/cardano/integrate-cardano-samples
```

and then for `cardano-cli` commands "reach up" for `.skeys` or `protocol.json`

```bash {3}
$ cardano-cli query utxo \
--testnet-magic 1097911063 \
--address $(cat ../keys/payment1.addr)
```

```bash
cardano-cli transaction sign \
--tx-body-file tx.draft \
--signing-key-file ../keys/payment1.skey \
--testnet-magic 1097911063 \
--out-file tx.signed
```

tweakch:
I think there should be a *Setup environment* page that guides can reference and explains the rational behind the structure...

katomm:
You are providing outstanding contributions here. I love the idea of doing a pwd before guides. 

Regarding directories, I am a little hesitant although I agree with you completely. For example, the "build with transaction metadata" was named "transaction metadata" in the beginning. Then later renamed for better user experience/SEO/marketing reasons. I'm afraid that might happen again and we'll always have to redirect the links.

tweakch:
Thank you I believe this portal is one of the most important things for driving adoption and I'm happy to help.

I can totally see why this is difficult to maintain.

Maybe we could embed everything in some kind of a curriculum with a number system that can be extended and has a logic to it. 

Like courses in university.

Then users could `cd` into `<courseId>` and work from there.

I'm makong this up as I go... 😅

tweakch:
As I re-read your comment I noticed a possible misunderstanding:

> I'm afraid that might happen again and we'll always have to ***redirect the links***.

[...]

PR #154, opened 2021-07-16
Add guide multi-witness transactions

tweakch:
I created a guide for multi-witness transactions

The reason for this guide is to show how one could build transactions with multiple inputs to a single output. 

It also talks about the stuff in the Explore Wallets guide a litte bit more and prepares devs for what's to come when smart contracts hit the public testnet. 

I also tried not to screw up with the paths in the `cardano-cli` but one should work through this guide first before merging. 

Finally I tried to incorporate a possible solution for issue #152 within this guide...


Mercurial (approved):
This is so awesome 👯‍♂️

katomm (approved):
Outstanding!

PR #204, opened 2021-07-30
CNFT.IO

itsmestale:
* Name: CNFT.IO
* Description: CNFT.IO Is the home to the Cardano NFT community, and now the home to the first NFT marketplace on cardano.
* Link:   https://cnft.io
* Tags:
  * `analytics`
  * `nft`

Added cnft.jpg and showcase data.


src/data/showcases.js
```diff
@@ -665,6 +665,15 @@ const Showcases = [
     source: null,
     tags: ["nft"],
   },
+  {
+  title: "CNFT.IO - NFT Marketplace on Cardano",
+    description:
+      "Mint, Launch, Buy and Sell Cardano NFTS on CNFT.IO. The first automated marketplace on Cardano.",
```
tweakch, line 671: Hi @itsmestale 
Please read the [About description section](https://github.com/cardano-foundation/developer-portal/blob/staging/CONTRIBUTING.md#a-word-about-project-descriptions) and change your description accordingly so we can approve your PR

tweakch (changes requested):

src/data/showcases.js
itsmestale: gotcha will do

tweakch (approved):

katomm (commented):
Hi, can you briefly elaborate on why the analytics tag was set? It doesn't seem to fit into this category. I'd suggest "token" and "nft".

itsmestale:
> Hi, can you briefly elaborate on why the analytics tag was set? It doesn't seem to fit into this category. I'd suggest "token" and "nft".

We just launched a few days ago, one of things I am working on atm are data and analytics. Market history, tracking price data on certain assets, trade volume etc. It's not there now but I thought it would be best to add the analytics tag as it'd be implemented pretty soon.

tweakch (changes requested):
I suggest you remove the `analytics` tag. This could be seen as misleading

MarcelKlammer (approved):

Mercurial (approved):

katomm (changes requested):
Sorry but the logo/screenshot still claims "Cardano's first CNFT Marketplace". 

Please also note that if people visit the site through the link they will receive a certificate error, because you link to www. but the certificate seems to be only for the domain without www.

katomm:
Closing due to inactivity [15 days] (If you think this is a mistake, the PR can also be reopened)

PR #212, opened 2021-08-03
Add tag/category feature for builder tools

katomm:
This PR adds the tag/category feature known from the showcase section also for the builder section. It extends the showcase component so that we can use it also to show a "get started" link.

I'd like to propose that "get started"-pages for operator tools should not go below the "get started"-category. Instead, we should use a new section in the "operate a stake pool" category: "operator tools". This is better for the overall user experience when searching and browsing categories; it also keeps things cleaner.

(I added a basic check to ensure the get started page will have a path below the operator category, but let's follow up on this when the first "get started"-page for an operator tool does a PR)


tweakch:
Great feature. I'll review this tonight

tweakch (approved):
Looks good :+1:

I've taken the liberty to add a small update to CONTRIBUTING.md

I agree we should move the Tools Section for user exp. (the guides feel a little out of place in Getting Started.) I'm curious to hear why you'd choose SPO section as parent

katomm:
@tweakch Sorry for being not clear enough. It should work like this: 
IF the builder tool has a get started page AND the tag "operator tool" THEN (and only then) the "get started"-page should go into the stake pool operator category. All other tools are unaffected.

rdlrt (approved):

gitmachtl (approved):

PR #220, opened 2021-08-06
Add Python Module

emesik:
This PR adds Python Module to Builder Tools page. All details are already in the source code.


rdlrt:
Thanks for the contribution, it seems the codebase being referred is skeleton at the moment for work you'll be doing. Having a section here would require you to have a fully functional release before it can be included.

emesik:
It is already usable module that might be used by developers right away. They are the target audience.

The project was intended to run on private wallets from the very beginning. It's not a public service, although it might be used to build such, as it provides useful tools that offer higher abstraction layer.

rdlrt:
> The project was intended to run on private wallets from the very beginning

Thanks for clarifying, in that case - can I request you to add "using cardano-wallet instance" to description?

Just some background on request : Currently there are a lot of projects trying to incorrectly use libraries they find online against API and end up starting to build public service using wallet backend (incl. some that are not supposed to be custodial). Having that clarity on description helps safeguard any future complains on text on developers.cardano.org.

rdlrt (approved):

katomm (changes requested):
Hello and thank you very much for the contributions.

Please allow me to request two changes:

- you must not set the "featured" flag yourself.
- you can not set the "getstarted" flag without having a "get started"-page in the portal. Please see the other builder tools with a get started page in the portal as an example, but we need to make this concept clearer in the future.

Thank you.

src/data/builder-tools.js
```diff
@@ -162,6 +162,14 @@ const Showcases = [
     getstarted: null,
     tags: ["featured", "operatortool"],
   },
+  {
+    title: "Python Module",
+    description: "The module provides tools for developers to accept and send transactions, manage staking and much more. It uses cardano-wallet as backend but is future-compatible with other solutions.",
+    preview: require("./builder-tools/cardano-python.png"),
+    website: "https://github.com/emesik/cardano-python",
+    getstarted: "https://cardano-python.readthedocs.io/en/latest/",
```
rdlrt, line 170: This should change to `null` as in example above

rdlrt (changes requested):

rdlrt (approved):

Mercurial (approved):

katomm (approved):
Thank you for your contributions @emesik!

Issue #125, opened 2021-07-13
include testnet-alonzo-genesis.json in "running cardano-node"


tweakch:
I was testing the [running cardano-node](https://developers.cardano.org/docs/get-started/running-cardano#configuration-files) section using the prebuilt binaries [build 6907168](https://hydra.iohk.io/build/6907168) and noticed the `testnet-alonzo-genesis.json` is missing from the list of configuration files [here](https://developers.cardano.org/docs/get-started/running-cardano#mainnet--production)

I will finish the tutorial on the website, fix the issues I see and submit a PR later

Mercurial:
Hi, the developer portal is intentionally not including any Testnet content as of right now. The goal of the developer portal is to showcase what you can do currently in the Mainnet.

PR #221, opened 2021-08-06
Missing Alonzo genesis files

devnullpal:

When I was runnng Cardano node for the first time on my computer, the Alonzo genesis files where missing. I downloaded them and the node ran successfully.


tweakch:
#125

rdlrt:
It is partially an issue with cardano-node CI/CD - which adds alonzo genesis for latest cardano-deployment , given latest is still marked to 1.27.0 for mainnet. I believe we need the change but as soon as new node release is marked.

Mercurial:
@tweakch and @rdlrt are both right that cardano mainnet only supports Mary era for now and therefore we are not adding genesis files for Alonzo just yet. But as soon as it hits mainnet then expect major updates regarding this into the developer portal.

PR #248, opened 2021-08-17
Create topo.png

avilsmeier:


rdlrt:
Please use single PR (#247) with both files

PR #267, opened 2021-08-30
installing-cardano-node.md Linux update

aaron-cesaro:

After some failed test I end up by installing the latest ghc version, instead of the suggested 8.1.4. 
In version 1.29.0 that compiler version prevent the creation of node.socket when starting the node for the first time.

I also updated all mentions from version 1.27.0 to 1.29.0

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -89,13 +89,6 @@ You can check if `ghcup` has been installed correctly by typing `ghcup --version
 The GHCup Haskell installer, version v0.1.14.1
 ```
 
-`ghcup` will install the latest stable version of `ghc`. However, as of the time of writing this, [Input-Output](https://iohk.io) recommends using `ghc 8.10.4`. So, we will use `ghcup` to install and switch to the required version. 
-
-```bash
-ghcup install ghc 8.10.4
-ghcup set ghc 8.10.4
-```
-
 Finally, we check if we have the correct `ghc` and `cabal` versions installed.
 
 Check `ghc` version: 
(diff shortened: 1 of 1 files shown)


docs/get-started/installing-cardano-node.md
```diff
@@ -89,23 +89,16 @@ You can check if `ghcup` has been installed correctly by typing `ghcup --version
...
 ```
 
-`ghcup` will install the latest stable version of `ghc`. However, as of the time of writing this, [Input-Output](https://iohk.io) recommends using `ghc 8.10.4`. So, we will use `ghcup` to install and switch to the required version. 
-
-```bash
-ghcup install ghc 8.10.4
-ghcup set ghc 8.10.4
-```
-
 Finally, we check if we have the correct `ghc` and `cabal` versions installed.
```
rdlrt, line 92: ghcup installer is preferred method to install ghc and manage different ghc versions.
Not sure why you removed this section

docs/get-started/installing-cardano-node.md
```diff
@@ -89,23 +89,16 @@ You can check if `ghcup` has been installed correctly by typing `ghcup --version
...
 
 Check `ghc` version: 
 ```bash
 ghc --version
 ```
 
-You should see something like this: 
+You should see something like this (current versione 8.10.7):
```
rdlrt, line 99: Unless advised otherwise 8.10.4 is the version recommended in docs, while higher than 8.10.7 'should' work, unless the baseline docs are updated, no reason to change the ghc version here

rdlrt (changes requested):
If you encountered an issue, it might be due to an error for which appropriate troubleshooting should be done, we should not change documentation for component versions from the ones mentioned in source repo of respective component based on single instance results (eg: using cabal update may update your dependencies that is often missed when upgrading)

rdlrt:
Closing as this is outdated (see #300 )

PR #275, opened 2021-09-05
iFrame is not big enough for Test Faucet

mikegcoleman:
#### Issue
Right now you cannot access the test faucet from https://developers.cardano.org/docs/integrate-cardano/testnet-faucet/ because the iframe isn't big enough. 

I've taken a guess at the correct size, and enable scrolling to address the issue (I can't fully test it because I don't know how to build the docs site locally).


katomm:
Thanks @mikegcoleman. To build the site locally please see https://developers.cardano.org/docs/portal-contribute/

katomm (changes requested):
Fix needs to be done in the css file, because the content changed slightly. I can take care of it, thanks for flagging.

katomm:
Fixed in #280

PR #276, opened 2021-09-05
Add explorer link and made them both English

mikegcoleman:
There is a link to the explorer above, but not actually in the step where you need to access the explorer. Added a link to make it a bit easier for users to follow along in case they missed the link previously. Also, I switched both links to the English landing page vs German since the tutorial is in English.

## Updating documentation
Added additional explorer link and changed both to the English version

On the [minting main tutorial](https://developers.cardano.org/docs/native-tokens/minting) there is a link to the explorer at the start of the "Check Your Node Status" section, but not actually in the step where you need to access the explorer. Added a link there to make it a bit easier for users to follow along in case they missed the link previously. Also, I switched both links to the English landing page vs German since the tutorial is in English.

### DIFF
--- a/docs/native-tokens/minting.md
+++ b/docs/native-tokens/minting.md
@@ -114,7 +114,7 @@ We will be using this technique of setting variables along the way to make it ea
 
 ### Check your node status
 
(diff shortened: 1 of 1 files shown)


rdlrt (approved):

katomm (approved):
👍

gufmar (approved):
Lgtm

PR #283, opened 2021-09-09
set placeholder fee for fee-calculating transaction

rphair:
@katomm has already suggested a PR be opened for this.

This is a response to a documentation problem reported [here](https://forum.cardano.org/t/official-tutorial-contains-a-joke-mistery/69572), saying that the fee itself must be known before the transaction that calculates the fee.  I believe this results from people running through the token creation exercises without yet having the basic transaction building experience (e.g. the process described [here](https://github.com/input-output-hk/cardano-node/blob/master/doc/stake-pool-operations/simple_transaction.md) for stake pool operators).

The forum comment implies setting `fee="0"` occurs so early in the flow of the affected page-- where it's included with general variables for the testing environment-- that it's easily missed.  I've moved it forward into the "Build the raw transaction" section so that the fee is initialised to zero immediately before that test transaction, along with an explanation of why the fee has to be specified twice (first an ignored value, then the calculated value).


docs/native-tokens/minting.md
```diff
@@ -285,9 +283,14 @@ $ txix="0"
 $ funds="1000000000"
 $ policyid=$(cat policy/policyID)
 ```
+Also, transactions only used to *calculate* fees still must have a fee set, even though this fee parameter is ignored. (The calculated fee will be included the second time the command is run: when the *real* transaction is built.)  So this variable must also be set:
+
+```bash
+$ fee="0"
```
AndrewWestberg, line 289: I would not set the fee to 0 even in the dummy transaction. Transaction costs are based on number of bytes and in the cbor of the final transaction, a 0 takes up fewer bytes than say 300000. I would set the value to 30000 when building the dummy transaction because it's likely more than what the transaction will cost and will take up the right amount of bytes in the dummy transaction. Then, when your final transaction says it's going to cost you 170000 lovelace, you can be happy.

AndrewWestberg: This answer comes from experience in building very large transactions with JorManager. Once your transaction becomes very large with lots of tokens and things, setting a 0 fee causes you to have the wrong answer and too low of a fee. It's hard to go wrong with the 300000.
https://bitbucket.org/muamw10/jormanager/src/c5d59c5164fae1fc44393c01f3062d32a13e0275/src/main/kotlin/com/swiftmako/jormanager/controllers/WalletController.kt#lines-266

rphair: thanks @AndrewWestberg - please see https://github.com/cardano-foundation/developer-portal/pull/283/commits/f249d430cad673c6eea6cbdf652798ac12e945f2

rphair:
I swear Github has promised in more than one place that it factors "dos" vs. "unix" line terminations.  Will try to submit again with fileformat closer to the original (which was effectively "dos"). 😓

katomm (approved):
Great contribution, also thanks to @AndrewWestberg

rdlrt (approved):

tweakch (approved):

den-is:
Hey guys, so is that related to assets minting transactions fees or is useful for any ADA transaction too, as a good new "best practice"? (for example p2p ADA transactions, rewards withdrawal)

PR #290, opened 2021-09-13
Update setting ~/.local/bin/ in PATH

peteski22:
---

When using WSL on Windows (with ZSH), following the [installation guide](https://developers.cardano.org/docs/get-started/installing-cardano-node/) leaves me unable to run the `cardano-cli` or `cardano-node` commands from the shell unless I am inside the `~/.local/bin` folder. Despite the `$PATH `appearing to include  the `~/.local/bin` folder from my home directory, I receive an error from ZSH that the command was not found. 

```
echo $PATH | grep /.local/bin
~/.local/bin/:...
```

I can successfully run the command inside the directory using:

```
cd ~/.local/bin
./cardano-cli --version
```

I believe the use of the `~` expansion in the export command in my shell profile is causing issues, and the fix is to replace `~/` with the `$HOME`. This produces the following output in the `$PATH` (where peteski is my username):

```
echo $PATH | grep /.local/bin
/home/peteski/.local/bin/:...
```

After restarting WSL or running `source ~/.zshrc` I am now able to reference the commands anywhere.

[Related information on tilde expansion](https://unix.stackexchange.com/questions/146671/does-always-equal-home/146697#146697)

#### Description

Changed reference to using `~/` to using `$HOME` for `$PATH` export of `~/.local/bin`

---

## Quickfix

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -210,7 +210,7 @@ cp -p "$(./scripts/bin-path.sh cardano-cli)" ~/.local/bin/
 We have to add this line below our shell profile so that the shell/terminal can recognize that `cardano-node` and `cardano-cli` are global commands. (`~/.zshrc` or `~/.bashrc` ***depending on the shell application you use***)
 
 ```bash
-export PATH="~/.local/bin/:$PATH"
+export PATH="$HOME/.local/bin/:$PATH"
 ```
 
 Once saved, reload your shell profile by typing `source ~/.zshrc` or `source ~/.bashrc` (***depending on the shell application you use***).


rdlrt:
Thanks for your PR.

While it  should be OK to use `$HOME` instead of `~` for wider support, I would think if we make the switch it should be consistent and all references for tilde should be replaced in that case

peteski22:
> 
> 
> Thanks for your PR.
> 
> While it should be OK to use `$HOME` instead of `~` for wider support, I would think if we make the switch it should be consistent and all references for tilde should be replaced in that case

I understand what you mean, the other 'fix' I think is to remove the quotes from the line.

`export PATH=~/.local/bin/:$PATH`

That is actually probably a better way to go about it, what do you think?

rdlrt:
> the other 'fix' I think is to remove the quotes from the line.

Having double quote prevents hitting issues with space/gobling if a folder in path has a space 🙂 see [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) for details

rdlrt:
Closing as subset of #341

PR #293, opened 2021-09-14
Notes on versions of compiler, random and wallet

erik-ros-devilsberg:

I've added notes to references of the compiler, random, and wallet versions used in this article as they are outdated. Ideally, the code would be changed so that these instructions work regardless of the version of the code that is current. These changes are beyond my current level of involvement with the project


katomm:
I understand why you are suggesting this. But the message itself will (likely) not be very helpful for new people. (also just because there is already a note box above it).

The focus should really be to bring the guides to node version 1.29.

PR #312, opened 2021-09-21
fix broken faucet link

vladanghene:


katomm (changes requested):
Hello, 
what exactly is the issue here? The link to https://developers.cardano.org/docs/integrate-cardano/testnet-faucet seems to work fine.

vladanghene:
yes well the href on that anchor is set to go above one dir `../` when in fact what you set as an example of a working link in your reply was from the dir `docs/getting-started/` which indeed works. I tried setting it as an absolute url to the first google result of a query `test ada faucet`, not having had delved into the root structure of the getting started app.

vladanghene:
this is where I was browsing from: https://developers.cardano.org/docs/get-started/testnets-and-devnets/

katomm:
The link is not broken. I guess the confusion is because you think "testnets-and-devnets" and  "testnet-faucet" are folders, but they are files. (testnets-and-devnets.md and testnet-faucet.md)

That means going from: 
/docs/get-started/testnets-and-devnets.md

To:
/docs/integrate-cardano/testnet-faucet.md

needs only one folder up.

vladanghene:
Okay, if you call [this Not Found page](working) then i guess it was my confusion

katomm:
I'm sorry @vladanghene, I'm reopening this.

katomm:
@vladanghene is it possible to provide the generated HTML code for that link? For me it is: 
Screenshot 2021-09-22 at 09 38 15

vladanghene:
yeap. for me too `href="/docs/get-started/integrate-cardano/testnet-faucet"`

i tested and it should be as I stated earlier, if you want it relative to the folder structure ``href="/get-started/integrate-cardano/testnet-faucet"`

katomm:
I'm able to reproduce this is now. While it works on Safari browsers it seems that Chromium forks add the "get-started" folder for some reason.

vladanghene:
i've been using safari from the get-go and it has never worked

katomm:
> seems like if I make it `../../` it doesn't pass the checks on your build tests.
Yes, because that would be wrong :) the cause here is something else.

vladanghene:
the cause is having `.../docs/..` in the href, it should have `/getting-started/..`

katomm:
I wish it were that simple. The markdown is:
`[Cardano testnet faucet](../integrate-cardano/testnet-faucet)`

(This is correct because we are in the "get-started" folder. This is also the reason why I closed your PR wrongly.)

vladanghene:
ok so i see now, the md is wrong having the jump to a folder above with `../`
inspecting other links 2 paragraphs above, I saw that the markdown you wrote simply binds `/docs/whatever` as if the homeurl is `https://developers.cardano.org/` and not `/docs/getting-started/`.

let me commit it as `docs/integrate-cardano/testnet-faucet/`

katomm:
While this will work I'd advise against /docs/whatever URLs because only relative links will be checked and flagged during the build of the site. (I know that there are still links with /docs/whatever, but we should avoid them.)

vladanghene:
been looking for a relative link with the structure you want `../` to successfully jump out of `get-started` and resume the href from `docs/...` and it seems the one I see in community page related to `if you want to join the [Plutus Pioneers](../smart-contracts/plutus#get-started-with-the-plutus-pioneer-program).` works fine ... it jumps out of get-started.
this one here does not however ...

i think you guys should rewrite the test checks so it includes `base` references in md as well as `/docs` hrefs... other than that, I have no idea why in some place it works and in others it doesn't. It's definitely not the [first time](https://stackoverflow.com/questions/7653483/github-relative-link-in-markdown-file) this is mentioned in regards to github's md.

rdlrt:
Looks like the issue is with docusaurus itself and how it routes between pages depending on browser's SVA context.
The markdown in it's current state should work fine if the corresponding fix for docusaurus is added.
The "fix" (essentially option to forcefully add `trailingSlash` which causes different formation of URL is [merged on their master](https://github.com/facebook/docusaurus/pull/4908), but a version is not yet finalized

katomm:
Thanks for the insights @rdlrt and thanks to @vladanghene for being persistent :) 

I'd suggest replacing URLs that are affected by this issue with the absolute "/docs/whatever" as an interim solution and then wait for a final release tag of docusaurus to address this.

vladanghene:
ok replaced it as interim solution, ready to be merged.

katomm (approved):
👍

rdlrt (approved):

PR #314, opened 2021-09-22
minor bug

gitmachtl:
added the 0x prefix, which is needed for the CLI to recognize the string as a byte array


rdlrt (approved):

katomm (approved):
Thanks Martin

PR #322, opened 2021-09-24
Change witness count to 2 for minting transactions

brouwerQ:
You sign with payment key and policy key, so I guess the witness count should be 2 for those transactions instead of 1?


katomm:
I need to summon @AndrewWestberg  and @ruttkowa for this PR. (see also #323)

katomm:
Sorry, opening this up again, as it is a different file. Thanks @brouwerQ

PR #323, opened 2021-09-24
Change witness count to 2 for minting transactions

brouwerQ:
You sign with payment key and policy key, so I guess the witness count should be 2 for those transactions instead of 1?

### DIFF
--- a/docs/native-tokens/minting-nfts.md
+++ b/docs/native-tokens/minting-nfts.md
@@ -388,7 +388,7 @@ As with every other transaction, we need to calculate the fee and the output and
 
 Use this command to set the <i>$fee</i> variable.
 ```bash
-fee=$(cardano-cli transaction calculate-min-fee --tx-body-file matx.raw --tx-in-count 1 --tx-out-count 1 --witness-count 1 --mainnet --protocol-params-file protocol.json | cut -d " " -f1)
+fee=$(cardano-cli transaction calculate-min-fee --tx-body-file matx.raw --tx-in-count 1 --tx-out-count 1 --witness-count 2 --mainnet --protocol-params-file protocol.json | cut -d " " -f1)
 ```
 
 And this command calculates the correct value for <i>$output</i>.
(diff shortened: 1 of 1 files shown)


katomm:
I need to summon @AndrewWestberg  and @ruttkowa for this PR.

docs/native-tokens/minting-nfts.md
```diff
@@ -388,7 +388,7 @@ As with every other transaction, we need to calculate the fee and the output and
 
 Use this command to set the <i>$fee</i> variable.
 ```bash
-fee=$(cardano-cli transaction calculate-min-fee --tx-body-file matx.raw --tx-in-count 1 --tx-out-count 1 --witness-count 1 --mainnet --protocol-params-file protocol.json | cut -d " " -f1)
+fee=$(cardano-cli transaction calculate-min-fee --tx-body-file matx.raw --tx-in-count 1 --tx-out-count 1 --witness-count 2 --mainnet --protocol-params-file protocol.json | cut -d " " -f1)
```
AndrewWestberg, line 391: 2 is correct for the `--witness-count` parameter when minting a token. However, `--tx-in-count` and `--tx-out-count` can be 0 I believe. In the code, they are used to add additional bytes to the fee calculation, but those bytes are actually already captured inside the matx.raw file. So, making those values 1 just makes you pay extra fees.  As long as you set something like 300000 for the fee when building the dummy transaction, setting `--tx-in-count` and `--tx-out-count` to 0 should work.

ruttkowa (approved):
Although I'm not that deep into which bytes get set under which conditions, at least for clarity and preventing misunderstandings, this is much better.

brouwerQ:
@katomm I don't know about that, I guess what he saying would be right. But if it's set to 0, it should be explained why I think or it'll be still confusing.

Also, you've closed #322 as a duplicate, but that's the same issue on a different file. Something got messed up while committing...

PR #329, opened 2021-09-26
Update installing-cardano-node.md MacOS update

jjjjjenkins:

`export PATH="~/.local/bin/:$PATH"` didn't work with me on Mac OSX 11.5.2 running zsh, since my .zshrc didn't recognize `~` as $HOME. Instead, it worked with `export PATH="$HOME/.local/bin/:$PATH"` which I believe should work for others as well.

An alternative update to instructions could be to specify that `export PATH="$HOME/.local/bin/:$PATH"` could work if `export PATH="~/.local/bin/:$PATH"` does not.

So it saves time for anyone debugging why running `cardano-cli --version` doesn't work.


rdlrt:
Duplicate of #290

PR #317, opened 2021-09-22
Changing absolute user to relative $USER path

Man7hano:
#### Issue

Fixes #316

#### Description of the change

I changed the absolute user path to $USER. Like so you can easily copy paste the commands from the documentation to your shell without any changes and no mix ups. I had the issue within the last export statement export CARDANO_NODE_SOCKET_PATH="/home/user/cardano/db/node.socket". I didn't change the user and couldn't connect to the node socket. It definitelly makes our lives easier if we can just copy paste it :).

PR #341, opened 2021-10-01
Update node checkout references, replace home directory references

rdlrt:
👋👋 Hello there! Welcome. Please follow the steps below to tell us about your contribution.

---

- Exclude specific version info in node compilation instructions
- Remove libsodium flag addition to cabal.project.local (it should only be required when falling back on system libsodium instead of fork, which as per the instructions in place - should not)
- Specify mainnet resource consumption may be higher than minimum specified
- Replace mentions of `~` and `/home/user` with $HOME across entire docs (replaces #317 and #290, but also includes other pages)

(PS: While stakepool handbook is entirely stale, references were only updated in those due to sed conditions 🙂 )

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -31,6 +31,10 @@ To set up the components, you will need:
 * A **CPU** with at least **two** cores
 * **8GB** of RAM and at least **10GB** of free disk space
 
+:::note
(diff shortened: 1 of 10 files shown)


gufmar (approved):
Lgtm

katomm (approved):
👍

PR #351, opened 2021-10-08
Add Pergamon Agora

kiriakos:
## Showcase addition

* Name: Pergamon Agora
* Description: An on-chain auction house for Cardano NFTs
* Link:   https://beta.pergamon.app
* Source: 
* Tags:
  * `metadata`
  * `nativetoken`
  * `nft`


kiriakos:
Hi there @katomm, 
1. I cannot understand what you mean with "traction" would you mind elaborating? There is no indication that contributions need to do anything in terms of `traction` 
2. This is a dApp, not a token. What made you think we are talking about a token?

PR #356, opened 2021-10-13
Update minting.md

arielelkin:
macOS's default Zsh shell, passing variables as command options doesn't work:

```
$ testnet="testnet-magic 1097911063"
$ cardano-cli query tip --$testnet
Invalid option `--testnet-magic 1097911063'
```

See:
https://discord.com/channels/386191477333557248/397539471702360084/897119938815336469


katomm (approved):
Thank you @arielelkin

PR #361, opened 2021-10-15
Update installing-cardano-node.md

extramileit:

Update the bash script code from 'cabal build all' to 'cabal build cardano-node cardano-cli'.  If you try to run `cabal build all` on latest tag of cardano-node code (i.e., 1.30.1), the build fails on tx-generator package with error message
```
cabal: Failed to build tx-generator-1.29 (which is required by
exe:tx-generator from tx-generator-1.29).
```
Since this documentation is specifically about building the cardano-node and cardano-cli, we can work around this by only building the specific executables needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ yarn-error.log*
 
 # markdownlint
 .markdownlint.json
+
+# intellij
+.idea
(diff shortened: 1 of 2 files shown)


katomm (approved):
Thank you @extramileit

PR #377, opened 2021-10-18
Update minting-nft to use transction build command

mzfr:
### Description of the change

Updated the `minting-nft` page. It was using the old `transaction build-raw` command. I feel that now `transaction build` works it's better to use that command. It helps to skip the fee calculation step.

Let me know if anything else has to be updated or changed.


docs/native-tokens/minting-nfts.md
```diff
@@ -370,46 +373,45 @@ echo $slotnumber
...
 --mint="$tokenamount $policyid.$tokenname" \
 --minting-script-file $script \
 --metadata-json-file metadata.json  \
 --invalid-hereafter $slotnumber \
+--witness-override 2 \
 --out-file matx.raw
 ```
 
-As with every other transaction, we need to calculate the fee and the output and save them in the corresponding variables (currently set to zero).
+If the above command generate an output similar to:
```
rdlrt, line 393: Reads incorrectly

mzfr:
@katomm I left the following comment:

> Actually, I added IF here because say the output value set above doesn't give any Minimum UTxO error then that would mean it worked and there is no issue.

Please let me know your thoughts on this. If you feel this should be changed, then I'll reword it :)

docs/native-tokens/minting-nfts.md
rdlrt: The suggestion was only for flow of the sentence (`If ... : ` felt like there is a continuation below, all good otherwise)

```suggestion
The above command may generate output as per below:
```

katomm:
Hey, @mzfr this was not about right or wrong. From my text feeling, I would go with @rdlrt's suggestion but I'm not a native speaker. I just wanted to know if there are more changes.

mzfr:
Oh okay, no problem. 
I'll make the suggested changes as soon as possible :)

mzfr:
@rdlrt @katomm I've made the suggested changes

rdlrt (approved):

katomm:
Hello, possible to resolve the conflict so that we can merge this? Thank you.

mzfr:
@katomm Conflict should be resolved now.

PR #382, opened 2021-10-20
Update minting-nfts.md

inspiraluna:
#### Issue

1. when burning a NFT the funds variable should be updated according to the current balance.
2. the slot variable is hard coded when re-running the transaction build which leads to errors for first time users and some confusion.

fixed typos in documentation


rdlrt (approved):

katomm (approved):
Thank you @rdlrt

PR #400, opened 2021-10-31
fixed 404 to min ada value calculations

nirhy:
clicking on the href for " Cardano ledger docs for further reading" led to 404
Updated the link to be https://cardano-ledger.readthedocs.io/en/latest/explanations/min-utxo-alonzo.html#example-min-ada-value-calculations-and-current-constants


rdlrt (approved):
Thanks

katomm (approved):
Good catch @nirhy

PR #404, opened 2021-11-07
Fix typo

dikaio:
_PR template removed_


katomm (approved):
Good catch @dikaio thanks

rdlrt (approved):

PR #405, opened 2021-11-08
Update testnets-and-devnets.md

Fallenstedt:
#### Issue
There is a missing link on this page here under `Smart Contracts`:https://developers.cardano.org/docs/get-started/testnets-and-devnets/#smart-contracts
I think the correct link should be this here: https://developers.cardano.org/docs/smart-contracts/

I updated the path which should go to this markdown file here: https://github.com/cardano-foundation/developer-portal/blob/staging/docs/smart-contracts/overview.md


rdlrt (approved):

katomm (approved):
Good catch @Fallenstedt thanks. Interestingly this was defined as a redirect but it seems to do not work.

PR #407, opened 2021-11-10
updated with testnet version

nilaysaha:
Not very obvious. But the earlier code does not work for testnet.


katomm (approved):
Thanks @nilaysaha

Issue #308, opened 2021-09-20
Bring the guides to node version 1.30.x (or latest)


katomm:
We really need to bring all the guides to node version 1.30.x (or latest). (#293)

os11k:
Hi! I just wanted to start work on this, but I was thinking about 1.30.1. Any particular reason to go with 1.29?

os11k:
Just to confirm, this page needs to be updated:
https://developers.cardano.org/docs/get-started/installing-cardano-node/
Right?

tweakch:
hi @os11k 

> Hi! I just wanted to start work on this, but I was thinking about 1.30.1. Any particular reason to go with 1.29?

I think there isn't. we can use 1.30 since it is the latest release. right @katomm?

> Just to confirm, this page needs to be updated: https://developers.cardano.org/docs/get-started/installing-cardano-node/ Right?

yes, ghc is at 0.1.17. if I remember correctly, so the listings are outdated. might make sense to use a similar approach as in #300 to keep the listings up to date

rdlrt:
> Just to confirm, this page needs to be updated: https://developers.cardano.org/docs/get-started/installing-cardano-node/ Right?

No, that page already refers to latest tag without hardcoded version references.

There are other sections that need to be bought in sync (Wallets, Operate a Pool, any CLI command arrears in minting guides)

katomm:
> I think there isn't. we can use 1.30 since it is the latest release. right @katomm?

Yes, 1.30.x (or latest) makes sense.

PR #414, opened 2021-11-20
Version update to 1.31.0 for binaries for "Installing cardano-node an…

os11k:

Updated links to binaries so it points to 1.31.0 for "Installing cardano-node and cardano-cli from source" page. Current manual contains links to 1.29.0 binaries, what are not supported at all at current point, so I think it is quite important to update at least till latest version for now.

---

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -15,10 +15,10 @@ This guide will show you how to compile and install the `cardano-node` and `card
 If you want to avoid compiling the binaries yourself, you can download the latest versions of `cardano-node` and `cardano-cli` from the links below.
 
 <HydraBuildList
-    latest="7408469"
-    linux="7408438"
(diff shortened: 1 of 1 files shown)


os11k:
Oops, seems versions are generated automatically. Need to check why those versions are incorrect. Maybe I will create an issue for that.

os11k:
Actually seems js code is checking what is latest version here "https://api.github.com/repos/input-output-hk/cardano-node/releases/latest" and then if we have not latest versions on the page we show small comments where to download latest code. As far as 1.29.0 should not be run anymore, due fork, it looks like it would be useful to update versions. What you think? Did I understood logic correctly? :)

katomm:
Yes, you are right. The guides need to be updated urgently.  #308

os11k:
I prepared separate pull request with more changes for this page, so I will close this one.

PR #420, opened 2021-11-23
Update installing-cardano-wallet.md for Linux

os11k:

Updated Linux installation manual for cardano-wallet. Now we will use stack for building cardano-wallet, what seems recommended way to do this it we look here: https://input-output-hk.github.io/cardano-wallet/developers/Building

### DIFF
--- a/docs/get-started/installing-cardano-wallet.md
+++ b/docs/get-started/installing-cardano-wallet.md
@@ -32,12 +32,87 @@ You must connect your `cardano-node` to the `testnet` network and make sure it i
 
 ### Choose your Platform
 
-* [MacOS / Linux](#macos--linux)
+* [Linux](#linux)
+* [MacOS](#macos)
 * [Windows](#windows)
 
-## MacOS / Linux
+## Linux
 
-In this section, we will walk you through the process of downloading, compiling and installing `cardano-wallet` into your **Linux / MacOS** based operating system. 
+In this section, we will walk you through the process of downloading, compiling and installing `cardano-wallet` into your **Linux** based operating system. 
+
+#### Downloading & Compiling
+
+We need to install stack, if we don't have it: 
+
+```bash
+curl -sSL https://get.haskellstack.org/ | sh
+```
+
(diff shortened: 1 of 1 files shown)


katomm (approved):
Thank you for updating this. I'd like to flag that technically this will work 1:1 for macOS. Only that this line:
cp -p ./lib/shelley/.stack-work/dist/x86_64-linux-*/Cabal-*/build/cardano-wallet/cardano-wallet $HOME/.local/bin/

Needs to be replaced with:
p -p ./lib/shelley/.stack-work/dist/x86_64-osx/Cabal-*/build/cardano-wallet/cardano-wallet $HOME/.local/bin/

Therefore I'm wondering if it still makes sense to split Linux/macOS this like you did.

os11k:
I do have mac near me and I even tried to compile it yesterday, unfortunately somehow it didn't went well. I'm still not aware what went wrong. Maybe later I will find time for this. I would assume same, that it should work as you proposed. Working with physical computers is pain in ass and rent mac in AWS is like 1000$ a hour. :(

katomm:
@os11k I have verified it myself. (macOs intel architecture)

os11k:
I will update pull request later this week so we will have MacOS & Linux same as it was before in one paragraph only with newest commands from me. If it worked for you, then I would expect that there no need to split it. I tested it on M1, maybe that why it didn't compile properly.

os11k:
Hi @katomm ! I updated as you proposed. Cheers!

katomm (approved):
@os11k awesome, thank you.

rdlrt (approved):
Approved as the dev-portal will be following respective repository's recommendation.

My only hesitation is at target itself (on `cardano-wallet` repo), as most other builder facing IOG components use `cabal` as recommended approach - but the wallet repo stands out with alternate instructions resulting in builders having to cater for different system requirements.

In ideal scenario, we could have both the versions - but that could be confusing to read (and in addition, if specific team does not include cabal as part of e2e testing - it might not be worth the additional effort on docs).

PR #428, opened 2021-12-01
Explain base16 encoding of token names

honungsburk:

In more recent versions of cardano-cli  the token names must be encoded in base16 for you to be able to create a transaction (I encountered this problem myself). With this PR the solution is made immediately available to the reader.


katomm (approved):
Great addition, thank you @honungsburk

PR #437, opened 2021-12-06
Prerequisites of build of cardano-node on ARM

JanRajtr:
#### Issue

On other platforms than x86/x64 (eg. ARM) it might be needed additional steps for successful installation and build of cardano-node and cardano-cli.

I followed [this post in forum](https://forum.cardano.org/t/cabal-fail-to-build-cardano-cli-on-utm/72837/5), this means I added additional steps in the documentation. The steps were tested on Ampere based VPS, Ubuntu 20.04.3 LTS.

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -209,7 +209,7 @@ We explicitly use the `ghc` version that we installed earlier. This avoids defau
 cabal configure --with-compiler=ghc-8.10.7
 ```
 
-If you are running non x86/x64 platform (eg. ARM, Apple M1...) please at this moment install and configure LLVM with:
+If you are running non x86/x64 platform (eg. ARM) please install and configure LLVM with:
 ```bash
 sudo apt install llvm-9
 sudo apt install clang-9 libnuma-dev
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -209,6 +209,16 @@ We explicitly use the `ghc` version that we installed earlier. This avoids defau
 cabal configure --with-compiler=ghc-8.10.7
 ```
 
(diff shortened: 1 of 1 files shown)


os11k:
Mac doesn't has apt command, so I would recommend to update this with proper mac commands

JanRajtr:
> Mac doesn't has apt command, so I would recommend to update this with proper mac commands

Good point, thank you. However mentioned forum thread afaik does not talk about solution for macOS - it is linux virtual machine on M1. Since I cannot test & confirm macOS M1 solution personally I removed mention about Apple from the committed paragraph.

katomm:
I don't have access to a M1 machine so I can't tell if it is enough to install the packages with Homebrew. Would be great if someone can validate this.

os11k:
I have M1 and I'm curios to test this, unfortunately I'm traveling now and my M1 is at home, when I will be back(in couple of week or so, if there will be no travel ban LOL) and if there will be still any needs to test, I might be interested in this challenge. Seems this is quite hot topic. :)

katomm:
Thank you @os11k for looking into this. (best to bookmark this, as we are merging this PR now)

Mercurial:
I just wanted to provide feedback that I was previously able to build Mac M1 with the existing guide https://developers.cardano.org/docs/get-started/installing-cardano-node#macos but not sure if something major has changed that it doesn't work anymore.

eastpiada:
The Armada Alliance has created an Apple Silicon M1 guide for getting a node running on a mini M1. It uses the build guide referenced above by @Mercurial which was successful for a 1.33 build. There are a few Armada Alliance specifics, but should be understandable by other non-AA pools.

[https://docs.armada-alliance.com/learn/stake-pool-guides/m1-native-macos-11-build](https://docs.armada-alliance.com/learn/stake-pool-guides/m1-native-macos-11-build)

PR #452, opened 2021-12-22
Add jq to the list of needed packages for CentOS

morucci:
Some parts of the documentation relies on the jq command that must
be installed. This package was missing on CentOS Stream 8.


katomm (approved):
👍

rdlrt (approved):

PR #458, opened 2021-12-31
extended-metadata implementation

cardanians:
extended-metadata implementation


katomm (commented):
I'll be blunt: this PR looks like a product placement. Instead, we can link to a CIP that's not promoting any specific product?

Any other opinions are welcome of course.

katomm:
Any more thoughts on this from anyone?

rdlrt:
I agree with the comment above, would like to see links to CIPs too

PR #460, opened 2022-01-03
Use stack install command to install cardano-wallet

morucci:
This change updates the documentation section for the cardano-wallet.
As the build instruction is based on the `stack` tool then it makes
sense use the `stack install` sub command to install produced binaries.

`stack install` copies binaries into $HOME/.local/bin/.
As of now copied binaries are: cardano-wallet, local-cluser and
mock-token-metadata-server.

By effect this commit removes the need to have a macos / Linux tabs
and thus simplifies this documentation section.

### DIFF
--- a/docs/get-started/installing-cardano-wallet.md
+++ b/docs/get-started/installing-cardano-wallet.md
@@ -90,35 +90,15 @@ We can now build `cardano-wallet` code to produce executable binaries.
 ```bash
 stack build --test --no-run-tests
 ```
+
 Install the newly built `cardano-wallet` binary to the `$HOME/.local/bin` directory:
-<Tabs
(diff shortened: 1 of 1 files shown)


katomm:
#420

katomm:
It would be good to have either a better description or more opinions here, as we seem to be going back and forth a bit between cabal and stack.

morucci:
> First I'd like to thank you for improving the documentation here. It would be good to have either a better description or more opinions here, as we seem to be going back and forth a bit between cabal and stack.

Thanks for the feedback, I've added a better commit message and PR message.

katomm (approved):
LGTM (tried only on macOS)

rdlrt (approved):

PR #461, opened 2022-01-04
update application link for the third cohort

benleelawrence:

Replaced old Plutus Pioneer Program application URL with the new cohort 3 link


katomm (approved):
Good catch

PR #462, opened 2022-01-04
chore(get-started): updating cardano-node requirements

mmahut:
## Updating documentation

Updating the minimal requirements for cardano-node, they have changed since the last edit.

https://github.com/input-output-hk/cardano-node/releases/tag/1.32.1

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -28,8 +28,8 @@ The components can be built and run on **Windows** and **MacOS**, but we recomme
 To set up the components, you will need:
 
 * **Windows**, **MacOS**, or **Linux** for your operating system
-* A **CPU** with at least **two** cores
-* **8GB** of RAM and at least **10GB** of free disk space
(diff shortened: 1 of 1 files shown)


katomm (approved):
👍

rdlrt:
> minimal requirements for cardano-node

Is that accurate? Those requirements should be for mainnet - we can run node with other networks on much smaller instances (hence, the note after the lines being edited).

Having said that, what we have now is "minimum requirement" either. The IO documentation does not differentiate between component requirements and network

os11k:
I tend to agree with @rdlrt , I do run pool in testnet with 8GB with no problem at all, so this will be misleading to put 12GB there.

katomm:
Could we agree on minimum requirements for different networks?  like @rdlrt suggested. I think that would be brilliant.

|Network          |CPU Cores* |RAM  |Storage |
|:---------------:|:---------:|:---:|:------:|
|Testnet          |2          |4GB  |10GB    |
|Staging          |2          |4GB  |10GB    |
|Mainnet          |2          |12GB |30GB    |

*An Intel or AMD x86 processor with **two or more cores, at 1.6GHz or faster** (2GHz or faster for a stake pool or relay)

PR #467, opened 2022-01-06
Exploring cardano-wallet - Some simplifications

morucci:
- Pipe "cardano-wallet recovery-phrase generate" into jq to get JSON array
- Do not recall POST payload in explanation
- Keep same wallet.id accross commands/examples
- Keep same address as in explanation
- Remove "--request GET" for curl command as this is the default


katomm (approved):
👍

PR #473, opened 2022-01-09
Update minting-nfts.md

Couder82:
## Quickfix

In metadada.json file the token name should be in utf-8 format, not in hex format, otherwise there is a problem with metadata.json displaying in pool.pm. 

This pull request tries to solve this problem. Feel free to change it to more understandable way, I just want to draw attention to this problem. 

Kind regards 
Martin from CBROS


katomm (approved):
Thank you.

PR #474, opened 2022-01-10
Modify Rust Library script 'replace' functions with regex

fill-the-fill:
Modify 'replace' functions with regex in order for functions to run more than once. 

Will definitely help avoid many bugs regarding naming in the future if more markdown documents will be added to the rust library.


katomm (approved):
LGTM

PR #483, opened 2022-01-17
Bump cabal version in Installing Node page

rdlrt:

Bump cabal version keeping up with updated upstream docs by IO to use 3.6.2.0


katomm (approved):
👍

PR #501, opened 2022-01-29
Fixing IFPS typos

plauche:

I noticed a few places in the _Minting NTFs_ page where _IPFS_ was spelled _IFPS_.


rdlrt (approved):
Thanks

katomm (approved):
Good catch, thanks

PR #508, opened 2022-02-02
Add PyCardano to builder tools

cffls:

Add PyCardano to builder-tools.js

PyCardano is a standalone Cardano client written in Python. The library allows users to build and sign transactions without depending on other Cardano serialization tools (such as cardano-cli and cardano-serialization-lib), making it a light-weight library that is easy and fast to set up in all kinds of environments.


rdlrt (commented):
Feedback:
Standalone Cardano client -> Would you consider renaming to standalone cardano library (as you're not connecting node via available mini-protocols but intermediate layer to query chain as backend, which is absolutely fine , and necessary given what's available from node as API, too - would ofcourse love to see project grow and achieve it's goals to be a node, you can always revisit the description accordingly)

cffls:
Thanks Priyank for spending time taking a look at this! Indeed, currently the library does depends on some intermediate layer to interact with the blockchain, and you are right that we will probably directly interact with node through mini-protocol in the future. I will remove the word "standalone" from this pull request accordingly.

rdlrt (approved):

PR #522, opened 2022-02-15
Updated with link to the testnet metadata explorer

shelley-dev:

Added link to our testnet transaction metadata explorer that provides many useful ways for developers to get easy access to the metadata posted to the testnet blockchain.
Developers can search for metadata containing specific text content and browse metadata by key or epoch.


rdlrt:
While it's a welcome addition, I am a bit confused if block explorer is the right place (as you'd typically expect explorers to provide interfaces to query blocks/transactions at minimum)

shelley-dev:
> While it's a welcome addition, I am a bit confused if block explorer is the right place (as you'd typically expect explorers to provide interfaces to query blocks/transactions at minimum)

See your point as the solution is not a block explorer. Simply reused the existing "chapter". Would you instead recommend creating a separate chapter called "Which metadata explorers can I use for the Cardano testnet?"

rdlrt:
Nah - just me pondering in terms of block explorer vs metadata transaction utility (for showcase tools), will wait for more opinions - happy to go with majority

katomm:
> "Which metadata explorers can I use for the Cardano testnet?"

I think in this case that would really be the best thing.

katomm (approved):
Thanks,  good addition to play with the testnet

rdlrt (approved):

PR #545, opened 2022-03-07
cleaned up commands,outputs

BharatAtEMURGOIO:

1. Cleaned up the xxd command to remove -b option (incompatible with -ps option) and unnecessary command option -c 80 as the output token name is less than 40 characters
2. Added env. variable for '--testnet-magic 1097911063' to help reduce clutter in the commands.
3. Corrected display of cardano-cli query utxo command as cardano-cli no longer shows the 'Testtoken'/'SecondTestToken' text, instead it displays the base16 encoded token names.
4. added some minor emphasis on certain text to highlight important points.


rdlrt (approved):

gufmar (approved):
LGTM

PR #566, opened 2022-03-26
documented a Create React App example under Serialization-Lib section in Get Started

dynamicstrategies:

Documented how to combine Create React App with the Serialization-Lib to get started quickly on Cardano. Targeted for those who already develop front-end app on other chains and especially for those that already use React JS. Coves use cases of sending ADA to interacting with plutus scripts. Documents how to clone and get started with the cardano-wallet-connector git repo


docs/get-started/cardano-serialization-lib/create-react-app.md
```diff
@@ -0,0 +1,300 @@
+---
+id: create-react-app
+sidebar_position: 2
```
katomm, line 3: This is a great addition. Could you please change the sidebar_position so that this example goes to the end of the current "Serialization-Lib" documentation?

katomm (changes requested):
Thank you.

katomm (approved):

docs/get-started/cardano-serialization-lib/create-react-app.md
katomm: We did that. :)) Open to better ways to address this if you want to share your thoughts.

PR #567, opened 2022-03-28
cardano-wallet: Stack no longer supported

sevanspowell:
## Updating documentation

- Update the cardano-wallet installation instructions to use Cabal instead of
Stack, as Stack is no longer supported by the wallet.
- Note that this only currently affects master, these instructions will not work with any released version of cardano-wallet, and the existing instructions hold for the most recent version. For this reason I think this PR should remain unmerged until the next release of cardano-wallet.

@katomm

### DIFF
--- a/docs/get-started/installing-cardano-wallet.md
+++ b/docs/get-started/installing-cardano-wallet.md
@@ -41,22 +41,7 @@ In this section, we will walk you through the process of downloading, compiling 
 
 #### Downloading & Compiling
 
-We need to install stack, if we don't have it: 
-
-```bash
-curl -sSL https://get.haskellstack.org/ | sh
-```
-
-Check the version that has been installed:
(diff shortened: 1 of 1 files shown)


rdlrt (approved):

katomm:
Thank you @sevanspowell. If this PR should remain unmerged until the next release of cardano-wallet then the best would be if you can convert it into "draft"-state.

katomm:
Hello @sevanspowell any update on when we should merge this?

rdlrt:
Since a release was marked yesterday, I believe we should be good to go ahead with the merge on this one

sevanspowell:
> Since a release was marked yesterday, I believe we should be good to go ahead with the merge on this one

Yes, thank you @rdlrt. That's true, we are good to merge this one.

PR #574, opened 2022-04-04
Add new showcase project Pavia

fill-the-fill:
* Name: Pavia
* Description:  Create, explore and trade in the first-ever Cardano virtual world owned by its users.,
* Link: https://www.pavia.io
* Tags:
  * `game`
  * `nftproject`


katomm (changes requested):
@fill-the-fill best to avoid "first-ever" something descriptions, this created some frustrations in the past here. 

Guildeline says: _Better describe what makes the project special, avoid phrases like "the first this and that". Granular details like which project was first is tribal attribute known to cause rift and conflicts._

katomm (approved):

PR #575, opened 2022-04-04
Add showcase project Atala Scan

fill-the-fill:
* Name: Atala Scan
* Description: Easy-to-use mobile app that gives customers real-time proof of products’ authenticity, recorded on the Cardano blockchain for life,
* Link: https://atalascan.io,
* Tags:
  * `analytics`


katomm (changes requested):
I would not agree with the analytic tag here. Suggestion: metadata, identity

katomm (changes requested):
Sorry for not being specific enough. Identity tag does not exist yet :)

rdlrt (approved):

PR #584, opened 2022-04-04
Add new builder-tools tag chain-index 

fill-the-fill:
- Add new builder tools tag
- Add chain-index tag to Oura project


rdlrt (approved):

katomm (changes requested):
Chain index is a good tag to have. I know we lack of a tag guideline but please let's stick to only lowercase tags and from my feeling the dash in the middle of the label should go away: 

- "chainindex" for the tag.
- "Chain Index" for the label.

(open for discussions of course)

katomm (approved):

PR #587, opened 2022-04-04
Add showcase Build On Cardano

fill-the-fill:
* Name: Build On Cardano
* Description:Discover 436 projects and dApps building on Cardano along with the developer tools you can use to build on Cardano. Explore similar projects and tools with ease and simply find you way back to where you started.
* Link:  https://builtoncardano.com
* Tags:
  * `ecosystem`


katomm (changes requested):
@fill-the-fill I recommend changing the description to "Discover projects", because "Discover 436 projects" will outdate quickly.

PR #605, opened 2022-05-03
cSnapshots Developer tool 

stakepool247:
## New Builder Tool
* Name: cSnapshots.io
* Description: Download the latest Cardano blockchain snapshot.
* Link:  https://csnapshots.io/
* Tags:
  * `pooltool`

PR #606, opened 2022-05-03
cSnapshots.io Showcase

stakepool247:
## Showcase addition

* Name: cSnapshots.io
* Description: Download the latest Cardano blockchain snapshot. We are updating the snapshots daily so you can sync your Cardano node much faster!
* Link:   https://cSnapshots.io/
* Source:
* Tags:
  * `pooltool`


katomm:
Closing this, because #605 is the same entry for the builder tools/operator tools.

PR #629, opened 2022-05-22
Add few more developer tools

KtorZ:

Adding a few more tools to the developer portal that are relatively mature and worth adding. Note that I have also:

- Added `kupo` and `scrolls` as _favorite_ because they're high-level API that are flexible and provide some lightweight/fast chain-index solutions for which they are none in favorite at the moment.
- I replaced `cardano-serialization-lib` with `cardano-multiplatform-lib` in the favorite, as the CSL is being 'deprecated' in favor of the latter. 

Cheers.


katomm (approved):
Awesome additions, much appreciated. 

(and I see what you did with the Rust Library Favorite, can we setup a vote on favorite developer tools?)

rdlrt (approved):

rdlrt:
> can we setup a vote on favorite developer tools

Ye - difficult to assess favourite tag candidates (imo it'd be one of each type, unless exceptional adoption - and prefer open source collaborative projects , especially when talking about builder tools/library 👍🏻 )

gufmar (approved):
Good additions

fill-the-fill (approved):

KtorZ:
> @katomm: can we setup a vote on favorite developer tools?
Sure. How do we proceed? 

> @rdlrt: imo it'd be one favourite of each type/use-case, unless there's exceptional adoption - and prefer open source collaborative projects

My thought also, hence why I am suggesting these two additions since there's currently a gap in this part of the spectrum.

rdlrt:
> If we do a vote I thought we could do it in the contribute category on Discord, but not everyone registers as a contributor to get access. (+ not everyone is a fan of Discord)
> 

We could do a poll on [github discussions](https://docs.github.com/en/discussions/collaborating-with-your-community-using-discussions/participating-in-a-discussion#creating-a-poll), tho any such polling is easy to spam - as our ecosystem has proven many times, a project can generate thousands of fake accounts to get listed and voted for a favorable outcome

KtorZ:
I think it's also somewhat fine to have a non democratic process for that, so long as the rationale for choosing the said projects is clearly established. Then, people are free to come and disagree if they feel like it but at least, the process is transparent and it doesn't feel completely arbitrary.

javiertc:
Maybe the favourites should be based on usage metrics like downloads or usage in other projects. For example, our cardano-wallet-js (https://www.npmjs.com/package/cardano-wallet-js) has always more than 250 weekly downloads on NPM, being one the most downloaded JS Cardano related SKD and it doesn't show as favourites here, and I think is a good candidate.

Also if a project is under heavy development and not production-ready like Scrolls or like Marlowe (that has been a favourite for ages) it shouldn't be on favourites. I believe this website serves better as onboarding to Cardano development and if you point newcomers to half-backed components we are basically increasing the learning curve for them (which is already quite high).

leobel:
> > can we setup a vote on favorite developer tools
> 
> Ye - difficult to assess favourite tag candidates (imo it'd be one favourite of each type/use-case, unless there's exceptional adoption - and prefer open source collaborative projects , especially when talking about builder tools/library 👍🏻 )

It'll be difficult (probably unfair) if we based it on people personal opinion, but if we instead used metrics then it won't. Also, I'm wondering why you want to restrict the number of favourite, and even to only one? I don't think it should be "only one winner" or based on "winner" at all. The important here is guide people to know the projects on Cardano and in some way "highlight the bests" right? Maybe establishing a threshold based on metrics, so any project reaching it is marked as favourite and not manually enforcing a fixed quota?

rdlrt:
> It'll be difficult (probably unfair) if we based it on people personal opinion, but if we instead used metrics then it won't. 

Used metrics/votes - especially in crypto world - are very  easily manipulated, besides. - they're very relevant to the package type, source used for deployment (not everyone uses things by downloading from a site, moreover - not every project is used in a way that'd have a metric, but will be used behind the scenes by many).

One favourite tag of each type is not a strict rule, neither a single person's opinion - but it's not meant to be too large to make it difficult to browse through.

KtorZ:
Note that I don't think the favorite tag should be exclusively reserved for production-ready tools or libraries. Precisely because it may be a good opportunity to bring attention to projects under development and seeking contributions / adoption.

Plus, a 'favorite' implies curation in my opinion. Thus, it's not something that should be based on vote or usage (that'd be 'most popular'). Favorite suggests that the tools are evaluated according to some subjective criteria. It may be a good idea to make those criteria public for transparency, but it's also not a requirement. Curators however should be disclosed, so that it's clear whose favorite we are talking about.

javiertc:
The issue there is that we end up with a list of Joe's favorites or Peter's favorites and at that point I would be better to post it on a personal blog post.

katomm:
> I'm just giving you an example of a tool that might be better for someone that is getting started and that is not even listed here 

@javiertc Thank you for your input on favorite tag. If you know a tool that is not listed here **please just add it**.

PR #646, opened 2022-05-29
golang cardano serialisation library

matiwinnetou:


rdlrt (approved):

src/data/builder-tools.js
```diff
@@ -272,6 +272,14 @@ const Showcases = [
     getstarted: null,
     tags: ["favorite", "library"],
   },
+
```
rdlrt, line 275: Would be good to remove this empty line for consistency

katomm (approved):
👍 (we should get rid of the requirement to sort the array alphabetically btw. Just adding at the end is fine as it is sorted automatically these days)

fill-the-fill (approved):

PR #649, opened 2022-05-30
Modify contribute button href to relative path

fill-the-fill:


katomm (commented):
Shouldn't this be "/docs/portal-contribute"?

katomm (approved):

PR #655, opened 2022-06-02
Cardano Failover Approach - Scripts and Docs

ResponsibleStaking:
## Builder Tools addition

* Name: *Simple Cardano Failover*
* Description: An approach to run a Hot Standby Block Producer Node with an easy to setup automated failover mechanism.
* Link:   [<link_to_project>](https://github.com/ResponsibleStaking/Cardano-Simple-Failover)
* Source: [<link_to_source_code>](https://github.com/ResponsibleStaking/Cardano-Simple-Failover/tree/main/script)
* Tags:
  * `operatortool`


rdlrt:
Thanks for the PR. Sorry but some concerns with your proposed scripts (it might be working for some atm but it'll gradually start failing more, especially for larger pools) :
1. switching firewall requires time to rebuild connections
2. with P2P-enabled nodes (in spite of against recommendations many have it enabled on mainnet) , this will no longer work , as you're not only reliant on incoming connections - so a firewall protection wouldn't be enough
3. As you noted the script does not check tip, it's a simple ping based wrapper
As such, we wouldn't want to recommend a solution that'd onboard unpredictable results -especially for larger pools (which will attract more connections naturally)

ResponsibleStaking:
Hi @rdlrt!

Thanks for your critical review!
Here is my feedback:
> 1. switching firewall requires time to rebuild connections
In the current setup (without P2P) rebuilding connections happens within seconds. Firewall based approach is used by some of the biggest pools I know about as well and I tested this multiple times to validate prompt reaction. Never took more than 5s to have new in connection from the relays again.

> 2\. with P2P-enabled nodes (in spite of against recommendations many have it enabled on mainnet) , this will no longer work , as you're not only reliant on incoming connections - so a firewall protection wouldn't be enough
I'm not 100% clear on the consequences of P2P for the BP nodes. I do not assume that they will be exposed to direct connections from other Nodes (except own Relays which are in between). So this should not be a problem. But the bi-directional communication could turn into a problem. I cannot achieve a state where the BP receives Blocks but does not send Blocks in this scenario. If IOG provides a flag to control block production (as they asked for in a previous survey) this would be the best alternative. I could state this precondition and give instruction to use a different "activation method" in P2P scenarios. Do you see alternatives to that which would work today already?

> 3\. As you noted the script does not check tip, it's a simple ping based wrapper
If I include a TIP check, would you consider it a qualified solution then?


> As such, we wouldn't want to recommend a solution that'd onboard unpredictable results -especially for larger pools (which will attract more connections naturally)
I purposely designed this failover approach in the simplest possible way (no additional infrastructure, just a simple service) to increase its adoption. Before I created a more complicated approach which was relying on a pro-active reporting of TIP to a microservice which then took failover decisions. Nobody was willing to use it because of it's complexity.

Please explain why you consider the approach to generate unpredictable results. I exactly wanted to avoid unpredictability by making it extremely simple to exactly avoid issues like e.g. unnecessary failover triggers based on too complicated health checks. In my scenario, the main node is never de-activated to also avoid scenarios where a false trigger leads to in-availability.

------
Bottom Line: Would you consider approving if I incorporate the following changes:

1. Add TIP Check 
2. Add a warning to not use it together with P2P enabled and comment that instructions for P2P will follow (alternative makeActive and makeStandby scripts)

rdlrt:
Thanks for the detailed response.

> In the current setup (without P2P) rebuilding connections happens within seconds

Without P2P - the reconnection would depend on the number of hot peers already on your topology, higher the number of active peers (and closer the last checkpoint), longer would be the wait times post a node being unresponsive/down.
However, with P2P - we're talking about a network with ~3000 registered pools, where beyond your topology, any outgoing peers can take precedence and get moved to cold peers (where 4 least connected peers are dropped at a specified frequency), unless specified to remain always hot.

> I cannot achieve a state where the BP receives Blocks but does not send Blocks in this scenario.

Correct - that is the concern (only when P2P is enabled). 

> If IOG provides a flag to control block production (as they asked for in a previous survey) this would be the best alternative.

Agree, it would actually be the only sane method (IMO) while other methods will end up to be hacks at best, and the guarantee to work will purely be on the SPOs comfort level with understanding config/networking (as they will need to be very specific) - the risk/downside being they'd end up creating adversarial forks if not done correctly. It's best to approach the chain of feature rather than having seperate solutions for P2P/non-P2P modes.

> If I include a TIP check, would you consider it a qualified solution then?

IMO - it would still remain a hack, just a more complete one 🙂

> I could state this precondition and give instruction to use a different "activation method" in P2P scenarios.

Running without P2P might not be a solution plausible in future, if the cluster of connectivity from P2P occurs outside of your topology's view.

> Do you see alternatives to that which would work today already?

IMO - you already have professional load balancers (eg: haproxy is a lightweight service used by many and is pretty straightforward to set-up once you have the configs sorted) to achieve this (which can not only check the liveness, but also the response against tip or have actions based on TCP flags) - even with P2P enabled. But I am not suggesting it's usage either, as I am not sure if P2P implementation is completely finalised - I did raise a [ticket at IO](https://github.com/input-output-hk/cardano-node/issues/1883) (tho that config was basic and old now) about viability, but did not receive any responses either (I believe for similar reasons as above).

To me it sounds like a no-brainer to chase for the feature request against node, and meanwhile - use leaderlogs or **temporary** solutions to ensure maintainance / outages are managed. The biggest concern in publishing these scripts is SPOs themselves, who may implement these incorrectly and end up creating adversarial forks on network

Ofcourse - that's just my individual opinion, FYI - that does not necessarily prevent your PR to get merged 🙂

katomm:
I would like to abstain here to make a review, but is there any update on this? thoughts or comments? Thanks @ResponsibleStaking

rphair (commented):
I also originally abstained from reviewing because @rdlrt had already said most of the things I would have said (based on 2+ years of SPO experience) and went into more technical detail about why not to do some things I'd only had the gut feeling not to do.

The author's last statements suggest that the best that could be done with this material is to fix some flaws with an approach which would still be highly specialised and not applicable to all SPOs.  Less technically qualified SPOs would be most drawn to the promise of "reliability" but would be the least likely to understand & fix anything that went wrong with this approach.

The term "Failover" as applied in devops implies both robustness of the failover solution (refuted by @rdlrt's comments) and a dependable consistency of the operating environment in the future (which, given the impending P2P Governor and other Cardano environmental feature, we won't have).

Hence it would imply to new stake pool operators that they're somehow obligated to employ this solution: otherwise they won't have "guaranteed uptime" or other marketable features which are of limited use given the fact that block production schedules are so well known in advance.

So I can't reject this addition but I neither would I want to suggest it to anyone else as an operator tool.  Like other projects which weren't voted into acceptance, I do think it's useful as a case study in Cardano devops, but I don't see how to qualify it anywhere else in the Developer Portal on that basis (@katomm?).

ResponsibleStaking:
Thanks for putting some attention on PR again :)
Since I created it I made some minor improvements to the script. 
The main difference is that the TIP of the master is also considered.
You can find the most current version of it here: https://github.com/ResponsibleStaking/Cardano-Simple-Failover/blob/main/script/checkStatus.sh

Anyways this is not affecting the main questions which you posted:

1. Will a failover be required in the future?
Probably not. The approach will require adaption once P2P is in use.

2. Does it make sense to recommend this to any SPO?
After running my Node for over a year (without any Downtime) it seems the risk of a node outage is relatively low (of course this depends heavily on the hosting).
On the other hand, applying the Script is not putting additional risk to the Master BP. The standby just starts when the master is not reachable. But there is a risk that both Nodes are running simultaneously in case of a misconfiguration.
So I'd recommend it to the rather technical SPOs. Originally I thought every SPO should be tech savvy which made me assume everyone should be trying to get a high availability setup. Obviously, this is not the case.

3. Is there a Place in the Developer Portal for this?
Maybe there could be a hint that there are HA Solutions available. Usage is optional and own research is required. Then my Github could be linked as an example instead of merging it into the Dev Portal itself. Of course, alternative approaches from other SPOs could be linked as well. Just as a through. If this does not make sense I'm also fine if you just decline the PR.

katomm:
Thank you for your contribution, this PR has been open for quite a while, and there seems to be no progress. We, therefore, close it, let us know with a comment in case you want to open it again.

PR #658, opened 2022-06-03
Update docusaurus.config.js

sanskys:
#### Issue

Fixes #640

Issue #641, opened 2022-05-27
SPO Sidebar navigation


sanskys:
I am updating the SPO part and will do it in several small steps. The first topic is:
-  to reorganize the Stake Pool section and integrate the content in the handbook into the course. The new sections could be 
1) Introduction 
2) Pre-requisites 
3) Installing & Configuring Cardano Nodes 
4) Registering a Pool 
5) Pool Maintenance 
6) Tools and Links

![SPO Sidebar](https://user-images.githubusercontent.com/80989645/170722739-45ef9d81-cc8d-4892-ab16-d188853f98c0.JPG)

katomm:
I think getting rid of the lesson titles and merging the "handbook" with the course is a great start. What we need to keep in mind is to set proper redirects for sites that no longer exist.  

This can be done in docusaurus.config.js below "redirects".

rdlrt:
Just a small point here, prolly good to avoid duplication (installing cardano node is already a section [here](https://developers.cardano.org/docs/get-started/installing-cardano-node) and could be referenced)

rphair:
Looking at SPO & developer security as an urgent, bare minimum update to the existing pages, I recently did a security review of the current material in the SPO & other sections: https://github.com/cardano-foundation/developer-portal/issues/42#issuecomment-1157180181 ... but I wouldn't be able to submit the changes I've started working on because they would no longer apply to this proposed page reorganisation.

Most of my work is related to key management which is also a topic here, so many of these changes might already be covered by someone else's pending work (?): https://github.com/cardano-foundation/developer-portal/issues/641#issuecomment-1147505973

@katomm @rdlrt what is the overall plan for coordinating & scheduling these changes?  I can still apply my audit results to a new set of pages (most of my security recommendations are already published [here](https://cosd.com/frankenwallet/prepare)) so rather than try to push it out this weekend I would prefer to wait for a sensible page reorganisation before adding the security related material.

In the meantime since building an Air Gap environment is necessary in general for key management (_by devs as well as_ SPOs) I'm going to work on that topic: https://github.com/cardano-foundation/developer-portal/issues/43#issuecomment-1157181551) ... this would be be pushed as its own page or set of pages in the Get Started section, so would be separate from the SPO menu structure.

Meanwhile I can wait for @sanskys @VitoMelchionna to finish this reorganisation before submitting any wide-reaching security update.  I'd also be happy to coordinate in the writing of some of the pages above & will watch here for a response about that. 😎

[...]

Issue #43, opened 2021-04-26
SPO course, security chapter missing


katomm:
The stake pool course needs a security chapter (better the topic should run through the whole course, but you have to start somewhere). We can use this issue to discuss ideas, requirements or any thoughts.

If anyone found their way here because they thought they were getting safety tips for their stake pool, please read on here: https://forum.cardano.org/c/staking-delegation/stake-pool-security/157

The security chapter should make clear that you must not have any payment key on the server. Also: how to handle keys in a secure way.

It would also be interesting to mention what people can do if there is more than one operator. Maybe one wants to share the stake key.
How to create, handle and sign transactions on an air-gapped machine.
Probably interesting also the scripts and how-tos that Martin did: https://forum.cardano.org/t/new-ledger-trezor-update-for-owner-pledge-stakepool-operator-scripts/43953

In general, the https://forum.cardano.org/c/staking-delegation/stake-pool-security/157 is full of great posts around this topic and I'm sure other people have even better ideas what to put in.

Please also see this related issue: https://github.com/cardano-foundation/developer-portal/issues/42

rphair:
@katomm we've prepared building instructions for an air gapped environment ([here](https://cosd.com/frankenwallet), posted on IO's Essential Cardano list under [Other Community Resources](https://github.com/input-output-hk/essential-cardano/blob/main/essential-cardano-list.md#other-community-resources)).  This would provide some foundation for a security topic, but where in the current Developer Portal outline would we contribute it?

1. Given that this documentation is already available externally, would it be enough to link to it, or must it be duplicated in a Developer Portal page?
2. Also, since an air gapped host is always useful to Operate a Stake Pool but also to Minting Native Assets, how would this reference best be posted so it's accessible to these two separate categories?

I know the security topic is broad, highly debatable, and full of liability issues.  The material I've written introduces some of the most critical security concepts so that the operator / developer can make their own decisions about them, and offers some practical guidelines.  To avoid over-committing to specific procedures, it could therefore be preferable to keep it as an external link (with the appropriate disclaimer) rather than reposting the material here.

These instructions were written mainly for building an air gap environment on a memory stick but the procedures & security concepts will also be applicable when installing a conventional air gap machine.

[...]

Issue #42, opened 2021-04-26
SPO course, security review needed 


katomm:
The content currently in the stake pool course needs a security review on best practices when it comes to key-related things.

We should not write content like "in the testnet, you can do it like this" - no, stake pool operators should learn how to do things the right away from the beginning.

Please also see this related issue: https://github.com/cardano-foundation/developer-portal/issues/43

rphair:
In trying to answer https://github.com/cardano-foundation/developer-portal/issues/43 I had to do a complete security review.  After a few days I have the results outlined & will get to work now on modifying the doc pages, and post it as a PR ASAP in the next few days.  Here are the main changes:

---------------

first, a new Air Gap Environment page in the Get Started section (not the SPO Course, since it's also needed for Devs to work with *their* keys).  As necessary, particular notes about working with keys & data will point back to this document.

interjections into several each of the Dev and SPO pages highlighting how to keep & work with the keys, rather than keeping them (as per the testnet oriented instructions) by default assumption on a node or Internet connected machine.

create another page (before Builder Tools, "Using keys on application servers") with some general cautions about keeping payment keys on live servers if & when necessary to use those particular tools, which have wallet passphrases & private keys in the cleartext code in several places through this section of the Dev Portal.

... This page would probably be just a stub for now because the problem of key management on servers doing automatic transaction process is more general & complicated than the problem of user key management which needs to be addressed first.  Some application oriented devs can help fill out this page later (?).

the "Integrate Cardano" section is now written as a series of testnet tutorials, so to avoid ripping all that documentation apart I would add a final "Working on the Mainnet" page in that section of mainnet equivalents if they exist (not for everything: there's no "faucet", you can't give yourself funds so no point creating an extra CLI wallet, etc.)

... to avoid duplication of effort the Mainnet page will insist that the reader complete the Testnet tutorials and then return here for an explanation of how to do these things in production.

... since "Build with transaction metadata" and "Discover Native Tokens" are also introduced with testnet commands, each of these sections will need to be concluded with their own "Working on the Mainnet" page.

The Stake Pool Course (particularly the SPO Handbook; the Lessons are all video) needs the same treatment but has some missing & redundant sections as well:

... The KES key instructions are incorrect (only appropriate to testnet) as well as insecure (missing any standard for working with these or any other private keys) and in the wrong place (instead of before generating the op cert, they're in an appendix of "Assignments" for the testnet).

[...]

PR #665, opened 2022-06-23
new pages: Air Gap Environment, Secure Transaction Workflow

rphair:
Despite what I said here (https://github.com/cardano-foundation/developer-portal/issues/641#issuecomment-1158344312) I'm submitting this relative to the mainstream branch, so this can be merged promptly without having to wait for the other PR to be merged first.  I'm just pushing out what I feel are the most urgent security recommendations: those that could lead to people's pledges or operating accounts being stolen or their stake pools pirated.

Pages are located at these positions in the sidebar because:
* Air Gap precedes Secure Workflow because it's a dependency for signing transactions with private keys.
* Secure Workflow precedes Builder Tools because some of these tools (e.g. `serialization-lib`) require working with private keys.

If https://github.com/cardano-foundation/developer-portal/pull/658 gets merged while I'm working on this let me know & I can rebase it on the resulting merged `staging` branch if necessary.  I'll be watching this discussion thread closely in the days ahead.

related issues: https://github.com/cardano-foundation/developer-portal/issues/42#issuecomment-1157180181, https://github.com/cardano-foundation/developer-portal/issues/43#issuecomment-1157181551


sidebars.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
           "get-started/installing-cardano-wallet",
         ],
       },
+      "get-started/air-gap",
+      "get-started/secure-workflow",
```
rdlrt, line 16: IMO - the secure-workflow/air-gap might be better suited under best-practices (there was a placeholder created in #658 in [sidebar.js](https://github.com/sanskys/developer-portal/blob/9363bff6ea66f2b76cb1ded261fccc6f8ab92286/sidebars.js#L159) ), but may be if there are multiple pages - can be top-level folder called `Security Best Practices` with pages underneath?

docs/get-started/secure-workflow.md
```diff
@@ -0,0 +1,263 @@
...
+
+## FAQ
+
+### Why can't I use `cardano-cli transaction build`?
+
+This is a convenient command to avoid the ["change" (return UTxO)
+calculation](../get-started/secure-workflow#calculate-the-change-to-send-back-to-paymentaddr)
+which requires you to prepare a test transaction,
+estimate fees, and calculate a final value of the funds to be moved.
+Instead, `transaction build` sends back "change" to a designated address.
```
rdlrt, line 237: Looks like there were carriage return from your text editor added (might not be intentional?)

docs/get-started/air-gap.md
```diff
@@ -0,0 +1,369 @@
...
+Air Gap environment on a bootable USB drive instead.  You can then boot a
+computer from this drive to have access to your secure resources and `cardano-cli`
+while isolating that computer from the Internet as well as any malicious software that might be installed on that computer.
+
+This loosely documented configuration has been called the **Frankenwallet**, with
+separate instructions at this link which mostly follow the procedure above and include
+semantics for using your bootable USB environment in secure & blockchain
+workflow:
+
+* **[The Frankenwallet](https://cosd.com/frankenwallet)**
```
rdlrt, line 358: Might be better to add a page explaining different address types within dev portal instead

rdlrt (commented):
Nice work overall, couple of comments added. Also :
1. Looks like there were carriage return from your text editor added (might not be intentional?)
2. `yarn.lock` file got added to the PR

sidebars.js
rphair: `stake-pool-guide/best-practices` in that pending PR is under the Stake Pool Guide hierarchy, but both of these pages are applicable to anyone (not just SPOs) who creates transactions manually: developers issuing minting transactions, etc.  To put these pages in any Stake Pool hierarchy would cause developers in general and token minters to overlook it or not use it.

I'm OK with putting these pages into a "Security Best Practices" section outside the body of SPO material.  Maybe keep the pages where they are now in linear order, and just put them under a `Security Best Practices` heading?

docs/get-started/air-gap.md
rphair: I don't understand how the suggestion relates to this part of the content.  That [Option 2](https://github.com/rphair/developer-portal/blob/air-gap%2Bsecure-workflow/docs/get-started/air-gap.md#option-2-install-your-air-gap-environment-on-a-persistent-usb-drive) section is there so readers know they can use Cardano securely without a second machine, i.e. that it's at least possible (even if they don't choose this method) to build a secure transaction hosting environment on a USB device instead.  What does that have to do with "different address types"?

rphair:
thanks; I swear I told it to ignore `yarn.lock` when I was doing the `git push` but I must not have understood the response 🤪 ... let's see if I can get rid of it now.  Like I said on the Discord I am not a real whiz with git/github.

I found a Git procedure about end-of-line normalisation but didn't find any standard for the Dev Portal suggested.  My 2 new files were created from scratch with `vi` on Linux so there's no reason CR would be added before LF by default.... but I did copy in part of another page (the transaction dialogue) from another page so this may not been consistent.  I'll go over it again but please in the meantime let me know what standard for line endings you want on the Dev Portal (LF, or CR-LF... i.e. Unix or DOS respectively) and I'll make sure this new material is consistent 😎

sidebars.js
rphair: p.s. OK I can see that is exactly what you are suggesting 🤪 ... will push it out that way & see how it works 😎

rphair: This is done & I think it looks/works better this way.

rphair:
> yarn.lock file got added to the PR

Deleted in last commit. 👍

> Looks like there were carriage return from your text editor added (might not be intentional?)

Please @rdlrt let me know how & where you are seeing these.  There are no carriage returns anywhere in any of the text files that I pushed from.  The "Git procedure" I mentioned was here ([Global settings for line endings](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings)) and there is no setting like that in my git config.

rphair:
thanks, I know what you mean now.  My `vi` based IDE makes it hard to work with text lines of unlimited length, since you can't reposition the cursor with a mouse.  Usually I just bang long lines through `fmt` so it wraps them at 72 characters.  If you give me a standard wrap length I'll run the markdown text to break it uniformly & re-push.  I agree it looks better when a consistent line length when viewing the page source.

docs/get-started/air-gap.md
rdlrt: My bad - I was a bit hasty with this comment (as community came up with the term `franken addresses` recently, which is unrelated to what you mention)

rphair:
imposing a limit on line length limits was unworkable since minor changes to text lead to multi-line differences in the diffs: impractical in the long run.  Looking around on other Dev Portal files, most lines are actually unlimited in length.  Markdown editors might also work this way but I've never found one on Linux (or anywhere else)

Anyway I found this to fold the text, heading & bulleted list lines (this is a general document formatter & translator):

https://pandoc.org/MANUAL.html (on Linux, `apt install pandoc`)

usage for this case: `pandoc -f gfm -t gfm -o air-gap-clean.md --reference-links --wrap=none air-gap.md`

Note you have to remove the Docusaurus header & merge it again with the cleaned-up markdown afterward, because `pandoc` doesn't recognise headers like this & folds them like regular text ("admonitions" are untouched).

Of course this changes almost every line of the file so it would make sense for any proofreaders to go through it all again.  When I was doing so I made some edits for spelling, clarity, etc. but they don't change anything substantial.

rdlrt (approved):
approving as a decent first draft, where-in a lot of good effort went in from @rphair , and corrections if any can be added in future as a PR once #658 is merged as well

fill-the-fill (approved):

katomm (approved):
👍

PR #686, opened 2022-07-08
Update Docusaurus to 2.0.0-beta.21

fill-the-fill:

PR #691, opened 2022-07-12
Docusaurus beta 21 built with yarn 3.x depends on prop-types

rphair:
New dependency (`prop-types`) created by https://github.com/cardano-foundation/developer-portal/pull/686.  Without it one gets the errors listed in https://github.com/cardano-foundation/developer-portal/pull/686#pullrequestreview-1034678501.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@popperjs/core": "^2.11.2",
     "clsx": "^1.1.1",
     "node-fetch": "^2.6.7",
+    "prop-types": "^15.8.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-popper": "^2.2.5",


fill-the-fill:
The error that is linked in [discord](https://discord.com/channels/875366566944968785/875615135035707412/996090562207813782) seems to refer to the sidebar issue that has been fixed.

Regarding the error code you have mentioned in [#686], could you please guide me through how did you get to that error?

katomm:
I did not look into this but could this be related to #693?

thenic95:
I'm able to build my up to date staging branch with `yarn build`

rphair:
I tested it again just now after merging https://github.com/cardano-foundation/developer-portal/pull/693 as well, with the same result (I did a similar test yesterday before this was merged).  At least on my software environment (Ubuntu 20.04 with usual dev tools), it even gets the same error with an unmodified clone of the repo:

```
$ mkdir NOT-MY-STUFF
$ cd NOT-MY-STUFF
$ node --version
v16.16.0
$ yarn --version
3.2.1
$ git clone https://github.com/cardano-foundation/developer-portal.git
Receiving objects: 100% (5727/5727), 153.53 MiB | 20.45 MiB/s, done.
Resolving deltas: 100% (3410/3410), done.
$ cd developer-portal
$ yarn install
... ➤ YN0000: Done with warnings in 37s 308ms
$ yarn build
...
Module not found: Error: @endiliey/react-ideal-image tried to access prop-types (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.
Required package: prop-types
...
```

thenic95:
> I tested it again just now after merging #693 as well, with the same result (I did a similar test yesterday before this was merged). At least on my software environment (Ubuntu 20.04 with usual dev tools), it even gets the same error with an unmodified clone of the repo:
>
> (**edit**) ... though in hindsight I can't be sure the dependency is coming from the new `@docusaurus` 🤔

I just did the same as you described here and for me it's working fine, I'm able to use `yarn build` without any errors

```
$ mkdir test
$ cd test 
$ git clone https://github.com/cardano-foundation/developer-portal.git
Receiving objects: 100% (5777/5777), 154.05 MiB | 12.10 MiB/s, done.
Resolving deltas: 100% (3485/3485), done.
$ node --version
v16.16.0
$ yarn --version
1.22.19
$ cd developer-portal 
$ yarn install
.yarn install v1.22.19
success Saved lockfile.
... Done in 32.12s.
$ yarn build
yarn run v1.22.19
✔ Client
✔ Server
  Compiled successfully in 39.48s
...
✔ Client
....
[SUCCESS] Generated static files in "build".
[INFO] Use `npm run serve` command to test your build locally.
✨  Done in 55.45s.
```

rphair:
It seems then that my version of `yarn` is too far ahead of what the rest of ye are using.  It must be that `yarn` v3 is much stricter about dependencies.  Next chance I get I'll back-rev somehow to `yarn` latest v1 and run through it again & post how that goes.

rphair:
With the new package environment of Docusaurus 2.0.0-beta.21 it does build properly with the legacy `yarn` 1.x (`1.22.19`).  So if we don't add the dependency above I'd only be concerned with what we would have to tell new, aspiring contributors to the portal.

As of recently, installation of `yarn` is recommended through `corepack` (https://yarnpkg.com/getting-started/install).  Currently this will install yarn `3.2.1` which as I've seen will fail our build process because the dependencies are calculated differently than in the legacy version.

So if this dependency can't be added, then some footnote (or a "known issue") will have to be added to the top level `README` to indicate what to do if & when contributors have trouble getting the project to build with the latest (stable or 3.x) version of yarn.

As I found today, once `corepack` is enabled and installs the latest `yarn` it's hard to eradicate.  The contributor would have to do one of these things:

disable & uninstall corepack, then install the legacy (1.x) version of `yarn` (through `npm` manually) that doesn't use the shim version in `corepack` or the one it copies into the user's `~/.yarn`.

(I didn't test this) set a global `yarnPath` as described here:

The `corepack` model seems to require that projects [set the `yarn` version explicitly](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines) in `package.json` (`"engines": {"yarn": "1.21"}`) which might be a better way of handling this.  But even if that were added to our `package.json` then contributors using `corepack` would still need to download that older yarn version to make it available (e.g. `yarn set version 1.22.19`)

Then once we decide how to best to accommodate both sets of contributors (those naturally on yarn version 1.x, and those on 3.x) we could update the README file telling them the best way to force themselves back onto the legacy version of yarn when building the portal. 😵‍💫

... or maybe if all this seems too much trouble we could simply add this package dependency, even if not required for building with the legacy `yarn`... so the portal builds without special instructions or other modifications on all versions.

katomm:
@rphair I'm also not able to reproduce this issue. If we can find at least one other person that runs into this, then we can just add the dependency. What machine are you running on? btw: are you using nvm to handle node versions?

rphair:
Some further notes: at least anyone on Ubuntu will have this problem, because:

1) The Dev Portal installation instructions stipulate `node.js` version >= 16.14 which is impossible to achieve from the Ubuntu repos alone.  Ffrom a fresh Ubuntu 22.04 (latest) installation (which you can test in a VM or AWS instance if you don't have Ubuntu on the desktop) it installs a much older version than the one we require:
```
# lsb_release -a
Release:    22.04
Codename:    jammy
# apt install npm
# node --version
v12.22.9
```

2) Therefore `node.js` v16 has to be installed as per [these instructions](https://github.com/nodesource/distributions#installation-instructions) from the repository https://deb.nodesource.com/node_16.x which, in turn, installs `node.js` 16.16.0, which includes `corepack`, which configures `yarn` to the current latest version `3.2.1`. 

i.e. this is not an exotic configuration: it's the one people will be most likely to be using if they have Linux on the desktop. 🧐

katomm:
@rphair I'm using yarn 1.22.4 but regarding node version: if you have multiple node projects, you probably won't get around nvm. (I wonder if we mention nvm in the install instructions) 

You can use exactly the versions the project requires, try it out with:
`nvm install 16.14 && nvm use 16`

rphair:
Nice to know about `nvm` but how `node.js` is installed doesn't matter in this case because (as it says in the [standard yarn installation page](https://yarnpkg.com/getting-started/install) already linked above):

> **Node.js >=16.10**
> Corepack is included by default with all Node.js installs, but is currently opt-in. To enable it, run the following command: `corepack enable`

If people are not supposed to use `corepack` (which currently installs `yarn` v3), this is one of the things they'd need to be told in the installation instructions, among several other things... including the apparent but undocumented consensus that the Developer Portal should only be built with `yarn` v1.

rphair:
The new mainstream version of Docusaurus doesn't have this dependency as considered by `yarn 3.2.2`, so https://github.com/cardano-foundation/developer-portal/pull/718 fixes this problem.  I can confirm that the current `staging` branch builds properly on both yarn v1 and v3.

BTW the default `yarn` version being set to `v3` happened because of a poorly documented step in the [yarn install instructions](https://yarnpkg.com/getting-started/install) which tells you to run `yarn init -2`... this installs the latest version of "Yarn 2" which is currently `yarn 3.x`.

But now that the dependency problem is fixed it won't make a difference which version of `yarn` contributors are using, so there's no need to address this in the portal instructions either & so this can be closed.  If controlling the version of `yarn` ever becomes an issue, we can use the notes here as a starting point.

katomm:
Thanks for the update. Good to know.

PR #692, opened 2022-07-12
Modify build scripts: Replace fs.rmdir with fs.rm

fill-the-fill:
In future versions of `Node.js`, `fs.rmdir(path, { recursive: true })` will be removed.


rphair (approved):
LGTM - https://nodejs.org/api/deprecations.html#DEP0147

rdlrt (approved):

katomm (approved):

PR #701, opened 2022-07-16
Add notes about the transaction era need to match

jfruiz11375:

I added notes about the transaction era must match. I was following the current documentation to learn the staking pool but since I am on the test-net the alonzo-era was giving an error when trying to submit the transaction build. I change the transaction build to use the --babbage-era and it was successful. So I just wanted to add notes in the documentation to help others as they take the stake pool course.

### DIFF
--- a/docs/stake-pool-course/handbook/register-stake-keys.md
+++ b/docs/stake-pool-course/handbook/register-stake-keys.md
@@ -72,9 +72,12 @@ It is either possible to simply keep the transaction as prepared above and send 
     > 997828515
 
 Herefore you would need to calculate the expression above and then build the transaction (as stated above) again, but adding the result calculated in the tx-out parameter:
-Note: In Babbage era the Babbage-era transaction is used by default and doesn't need to be specified on command line.
+
+:::note
+In Babbage era the Babbage-era transaction is used by default and doesn't need to be specified on command line.
 
 The transaction build command works only when transaction era is the same as network era. It cannot build e.g. Mary-era transaction on Babbage-era network. In that sense the --babbage-era, --alonzo-era, etc. options are redundant. The transaction build should figure out what the network era is and build same era transaction.
+:::
 
 ```sh
 cardano-cli transaction build \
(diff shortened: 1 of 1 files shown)


rphair (commented):
some observations:

1. although this material is going away in favour of pages in the new Stake Pool Guide (https://developers.cardano.org/docs/stake-pool-guide/) I wouldn't have a problem adding a note like this to the old material.
2. I am assuming the statement in the note is true since `transaction build` needs to connect to a live network and therefore it doesn't make sense to specify the ledger era.  We found this out the hard way over the course of this discussion: https://forum.cardano.org/t/please-use-cardano-cli-transaction-build-instead-of-cardano-cli-transaction-build-raw/94919/5
3. @rdlrt (or other) - before adding such a statement, do we have a way of verifying that it's strictly true under all circumstances?
4. @jfruiz11375 - after adding this note, why have you not also edited the command which follows to remove the `--alonzo-era` option, if it's not necessary?

jfruiz11375:
I am new here so sorry was not sure if I should change the code or not.

I was also thinking that all the material is old so maybe there was a
reason for that command to stay that way.

Should I go change the command?

rdlrt:
> although this material is going away in favour of pages in the new Stake Pool Guide

Personally I wouldn't be too big to update old guide pages, but put corresponding changes to updated page (for which it's easier if initial draft is ready , even if not fully ready quicker).

> before adding such a statement, do we have a way of verifying that it's strictly true 

It will be true with upcoming node versions (see [here](https://github.com/input-output-hk/cardano-node/pull/4135) ), tho it may be a good idea to switch to `build-raw` instead - which would allow defining valid transactions in lower era, especially if a feature is not essential.

rphair:
I've been in touch with the fellows who initiated https://github.com/cardano-foundation/developer-portal/issues/641 so I'll keep an eye out for it when those pages come out: both to use `build-raw` and to address the question of using a different ledger era (also cross referencing it now).

jfruiz11375:
Great, thanks!

rphair:
@jfruiz11375 are you therefore OK if we close this pull request... knowing it will be noted in the revised material coming out soon?

jfruiz11375:
Yes, go ahead.

Issue #680, opened 2022-07-04
Add new tag "DAO tooling" for showcase section


thenic95:
- DAOs are coming to Cardano and so are the tools needed to manage them
- Currently we don't have a dedicated "DAO tooling" tag in the showcase section to display those projects
- Should we add the tag "DAO tooling" to the available tags?

- The new tag would fit perfectly for this [PR](https://github.com/cardano-foundation/developer-portal/pull/674)

katomm:
Having this tag for #674 makes a lot of sense. I'd go with "DAO tool" instead of "DAO tooling". I'd suggest to remove "wallet".

katomm:
I'm closing this, because we did the tag and merged it. (in #677)

thenic95:
The tag hasn't been created yet. I still need to open a PR regarding this. Could we keep this issue open for now?

PR #702, opened 2022-07-20
feat: Add haltscam project

HT-Moh:
Haltscam allows and provides a means of crowd-sourcing risky Cardano addresses, allowing community members to contribute to the database and prevent their fellow adopters from falling victim to scams or solicitation.

## Showcase addition

* Name: Haltscam
* Description: Haltscam allows and provides a means of crowd-sourcing risky Cardano addresses, allowing community members to contribute to the database and prevent their fellow adopters from falling victim to scams or solicitation.
* Link:   https://haltscam.com/
* Source: null
* Tags: `ecosystem`
> note: I can not find the security tag.


rphair (commented):
thanks @HT-Moh for your submission.  There are many changes to the white space, but to me they all look like improvements (e.g. removing white space that never should have been there & setting indentations properly).

I'm OK with all of this as soon as appropriate tag(s) can be decided upon.  I wouldn't call it an `ecosystem` project because it doesn't provide a general view of the ecosystem as all the tools there do.  Please stay tuned to see if we have agreement about adding the project & then what tag to use 😎

rphair:
@katomm @fill-the-fill @rdlrt I've applied the `showcase` tag because to me this project looks useful and seems legitimate in every way that I can tell.  The crazy idea everybody should have, of course, is that the project itself could be a scam 🤣 but it's coming from the SPO of a CF delegated pool & is verifiable a few other ways.

So the only question would be the tag.  Since this tool is basically crowdsourced analytics to create a list of blacklisted addresses, my choice would be to give it the `analytics` tag.

If you don't agree it fits in there, perhaps instead create a new tag like `blacklist` (where it might remain alone for some time... I'd prefer "blacklist" because the term is more specific than "fraud" or "scam" even though these latter terms maybe easier to recognise.

HT-Moh:
@rphair thanks, I couldn't find a tag as well.

katomm:
@rphair You did the right thing, and good to know: the showcase label does not evaluate projects. It is simply there to categorize the PR. (We can use this for analytics to see for example how many PR on the portal are regarding to showcases or builder tools)

If a project needs a tag we can create a new. Personally I'm not sure about "blacklist", seems confusing to me without any explanation and ofc "scam" is misleading, people might think this is a scam😁. What with "scam-fighting" or similar?

rdlrt:
> What with "scam-fighting" or similar?

"safety" ?

thenic95:
The description of the "Analytics" tag checks out:

`Tools that provide special insights related to Cardano.`

I also like using the name **safety** for the possible new tag.

rphair:
I also like adding it to `analytics` where it helps round out that category, but if people are looking for a security or safety tool it won't be the first place where people will look.

Having `safety` on the tag bar would help avoid this project & someday others like it getting missed.
According to the precedent here (https://github.com/cardano-foundation/developer-portal/issues/680) more "safety" tools are likely "to be coming" and accordingly there should be a new tag.

katomm:
So how should we proceed here? "Analytics" and "safety" tags are fine but I think "ecosystem" tag is clearly something else if you compare this project to the others here: https://developers.cardano.org/showcase?tags=ecosystem

src/data/showcases.js
```diff
@@ -1082,7 +1082,16 @@ const Showcases = [
...
+    tags: ["nftsupport", "opensource", "token"],
+  },
+  {
+    title: "Haltscam",
+    description:
+      "Haltscam allows and provides a means of crowd-sourcing risky Cardano addresses, allowing community members to contribute to the database and prevent their fellow adopters from falling victim to scams or solicitation.",
+    preview: require("./showcase/haltscam.png"),
+    website: "https://haltscam.com/",
+    source: null,
+    tags: ["ecosystem"],
```
rphair, line 1094:
```suggestion
    tags: ["analytics"],
```

rphair (commented):
@HT-Moh since it's the simplest modification (without creating a new tag), can you just change the tag to `analytics`?  If any of the other reviewers feel strongly that a `safety` tag should be created and applied in this case, that's OK with me as well.

HT-Moh:
@rphair thanks, done.

rphair (approved):

rphair:
@HT-Moh we need a couple more green-check (accepting) reviews before it can be merged.  I'm inviting some more people explicitly with the privilege to do this... stay tuned 😎

fill-the-fill (approved):

rdlrt (approved):

HT-Moh:
@rphair Thank you

PR #703, opened 2022-07-20
Fix hover effect on buttons & make border thicker

fill-the-fill:
Seems like latest Docusaurus update broke the hover effect on all of the buttons

- Fix css of the buttons
- Increase border thickness of the buttons


katomm (approved):
LGTM

PR #707, opened 2022-07-25
Modify announcement bar to be closable

fill-the-fill:
Modify announcement bar to be closable

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -38,7 +38,7 @@ module.exports = {
         `The <strong>Cardano State of the Developer Ecosystem 2022</strong> survey is out! Your voice matter, please take a few minutes to <a target="_blank" href="https://cardanofoundation.outgrow.us/cardano-developer-ecosystem-2022">participate to the survey</a> 💗.`,
       backgroundColor: "#2AA18A",
       textColor: "#FFFFFF", // Use #FFFFFF
-      isCloseable: false,
+      isCloseable: true,
     },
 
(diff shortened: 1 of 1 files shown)


rphair (approved):
Just noting @KtorZ's comment that the survey is running through the 29th (4 more days from now) (https://github.com/cardano-foundation/developer-portal/pull/704#issue-1313064636) but does not say the bar _cannot be closable_ before then; only that it should remain there. 🤔

fill-the-fill:
> Just noting @KtorZ's comment that the survey is running through the 29th (4 more days from now) ([#704 (comment)](https://github.com/cardano-foundation/developer-portal/pull/704#issue-1313064636)) but does not say the bar _cannot be closable_ before then; only that it should remain there. 🤔

I believe in case someone already did the survey, it would be annoying to constantly see the announcement bar on top

rphair:
> in case someone already did the survey, it would be annoying to constantly see the announcement bar

I agree & have `display: none`'d web site components in browser style sheets for less... it would be a shame for people to have to do that to the Developer Portal 💔

rdlrt (approved):

KtorZ:
Note that the survey is ending tomorrow. So we might as well remove the bar 😅

katomm (approved):
Agree, they should always be closeable.

katomm:
> So we might as well remove the bar

Fine to remove a closeable bar. ☝️😁

PR #708, opened 2022-07-26
Fix broken links for Jobs at dcSpark and EMURGO

thenic95:


rphair (approved):
I hate to change the correctly spelled Emurgo URL to an incorrectly spelled one but I also know how futile it is to submit a company contact form saying something like "fix your web site" with about 100% chance of getting marked as spam. 🤪

Someone must have accidentally changed the page stub and could just as randomly notice the problem & change it back at any time: but it might be months for that to happen, and then we'd have to adjust again.  So it would be better to get them to correct the URL if anybody knows somebody there who can do that (@vsubhuman?)

thenic95:
> I hate to change the correctly spelled Emurgo URL to an incorrectly spelled one ...

I didn't even notice that, but yeah that hurts 😂

fill-the-fill:
> I hate to change the correctly spelled Emurgo URL to an incorrectly spelled one but I also know how futile it is to submit a company contact form saying something like "fix your web site" with about 100% chance of getting marked as spam. 🤪
> 
> Someone must have accidentally changed the page stub and could just as randomly notice the problem & change it back at any time: but it might be months for that to happen, and then we'd have to adjust again. So it would be better to get them to correct the URL if anybody knows somebody there who can do that (@vsubhuman?)

Maybe @mallapurbharat could help with this one

katomm (approved):

fill-the-fill (approved):

PR #718, opened 2022-08-01
Update Docusaurus to 2.0.1

katomm:


rphair:
this sounds great @katomm, looking forward to testing 😎 ... but why was this checked in with `yarn.lock` rather than modifying the relevant lines in `package.json`?

I know you might have your reasons & it's no problem for me... just need to know how to test accordingly (I thought we weren't checking in `yarn.lock` as per https://github.com/cardano-foundation/developer-portal/pull/665#pullrequestreview-1017825534).

Just need to be sure of this because I'm also looking forward to testing relative to the package configuration issues observed in https://github.com/cardano-foundation/developer-portal/pull/691 & would like to see if the dependencies are built properly with `yarn 3.x` now that Docusaurus is out of beta.

katomm:
> why was this checked in with `yarn.lock` rather than modifying the relevant lines in `package.json`?

My apologies.

katomm:
Okay I fixed it, @rphair. Thanks. yarn.lock should be handled only by the local yarn installation. (not sure if it makes sense to put it in git ignore, I guess not so we need to be careful on that)

gufmar (approved):
Lgtm

rphair (approved):

rdlrt (approved):

Issue #740, opened 2022-08-15
Showcase Section - Properly display new additions 


thenic95:
Hi,

I was just thinking that it might be cool to more prominently display new additions to the showcase section.

Is there a way we could automate that without using a dedicated tag where we wouldn't have to open a PR to remove it, once a given time has passed.

**Why would we need that?**

Right now, you can't really tell what projects were recently added if you're not checking out the PRs on GitHub.

I believe it would a lot of value and would be just a cool feature.

Cheers,
Nicolas

katomm:
I think this is a great suggestion. If we want to do this and assuming that people add projects at the end of the array (which is usually the case) we would need to look at the last x entries and list them with a dedicated category tag. That's all.

katomm:
nice poll, I mean PR: #749

PR #744, opened 2022-08-20
Rework 'Get Started: Cardano Components'

KtorZ:
Somehow, the presentation of the Cardano components felt completely
  off. Most 'components' presented in this section were actually just
  builder tools promoted as "core components". Truth is, they're not,
  they are merely builder tools that happened to be developed by the
  same company also developping core components.

  When it comes to Cardano's architecture and core components, we can
  roughly divide it in 4: settlement, consensus, networking and
  scripting − despite consensus and networking being heavily
  intertwined.

  Hence, I've reworded the components section to reflect that, and moved
  all builder tools where they belong. Flagging the `cardano-cli` as
  _favorite_ since it is a must to learn to use. Interestingly enough,
  I couldn't find any good "Getting Started" source for the cardano-cli,
  walking people through basic usage and commands that this tool
  provides. There are various guides and tutorials that make use of the
  command-line but I have yet seen any that really offers an
  introduction the the cli itself. Something to write?

image


src/data/builder-tools.js
```diff
@@ -180,6 +180,62 @@ export const Tags = {
...
+    title: "cardano-cli",
+    description: "The companion command-line to interact with a Cardano node, manipulate addresses or create transactions.",
+    preview: require("./builder-tools/cardano-cli.png"),
+    website: "https://github.com/input-output-hk/cardano-node/tree/master/cardano-cli#cardano-cli",
+    getstarted: null,
+    tags: ["favorite", "cli"]
+  },
+  {
+    title: "bech32",
+    description: "Convert to and from bech32 strings from the command-line. A simple and easy-to-use unix utility.",
```
rdlrt, line 193: Not sure if this deserved to be a seperate entry, as it's quite basic utility

rdlrt (approved):
LGTM, agree with all - except maybe reconsideration for having bech32 as a builder 'tool' (I dont have too strong feeling with/against - happy with as-is too)

katomm (approved):
👍👍

src/data/builder-tools.js
KtorZ: I tend to agree...  I only put it there because it was listed as one of the 'Cardano components'  😬

It's basic utility, probably doesn't deserve a spot in the builder tools but simply, mentioned somewhere in one of the guide as some useful tool to have on one's Swiss knife.

rphair (approved):
I'm OK having `bech32` listed as a builder tool.  When I began as an SPO I'd seen those addresses but had no idea about the system behind it... so I believe it will be helpful for newcomers to have an emphatic reference to it when they are still at the browsing stage.

thenic95:
Shouldn't we add new builder tools to the bottom of the existing list?

rphair:
@thenic95: 
> Shouldn't we add new builder tools to the bottom of the existing list?

Since none of these tools are new, doesn't it make the most sense for them to appear early in the list, if not the beginning (since these tools have been around since the beginning)?

KtorZ:
The list is rendered in alphabetical order anyway. So where we add stuff matters not.

katomm:
> The list is rendered in alphabetical order anyway. So where we add stuff matters not. 

I'd say because it **can** be  rendered in different ways (like alphabetical), we **should** add projects to the end. This would allow us other views like proposed in #740

fill-the-fill (approved):

rphair:
OK @katomm that's a good idea in general, and we're OK to merge this one specifically without needing to put new items at the end because none of these tools are new.

PR #745, opened 2022-08-22
Add cexplorer

cexplorer:
add cexplorer explorer with testnet support -  https://github.com/cexplorer/cexplorer.io/blob/main/doc/instances.md


rphair (commented):
@cexplorer can we -1- discuss with you how many categories this site could or should be in?  It seems to fit (at least):

* Showcase: Pool Tool / Native Token
* Builder Tools (as submitted in original PR): Operator Tool / API

If linking to the top level web site [cexplorer.io](https://cexplorer.io) I believe it should be added in all appropriate places to avoid confusion for the reader (@katomm maybe I'm wrong about that & if there is some precedent please say so).

-2- Because Cexplorer seems to be "the new" [adapools.org](https://adapools.org), it could be confusing to the community to have the old tool but not the new tool there currently in the Showcase under Pool Tools.  (This is what I was thinking when I added that label above but then removed it.)

Some messages e.g. the ADApools Telegram bot suggest the old site is being deprecated & could you please tell us if & when that might be happening, so we can judge how it would affect the listings here? i.e.

* Is there a deprecation plan for the older site?
* If ADApools is being kept, what will it still be used for if Cexplorer does all the same things?

-3- Probably a screen shot of the home page would suffice as a featured image even if Cexplorer is listed multiple places in both Builder Tools and the Showcase... but since it's allowed a different feature image in both, the Builder Tools image could be a view of the web site that illustrates the API or testnet support.

In any case @cexplorer can you plan to prepare 1 or 2 featured images, depending upon the categories & tags where we agree this should be listed?

katomm:
Looks like this is a rebrand of adaex.org (which we have already in showcase)

Best to change the old entry.

katomm:
I feel like we're stuck here Any updates @cexplorer? Should we merge this in and remove adapools? or do you want to change the adapools entry yourself? (because if this PR goes in then we have two projects in that are the same)

rphair (approved):
Created simple featured image based on logo & linked it to your added code.
Result (with my last commit) builds OK with current `staging` ✔️

rdlrt (approved):

katomm:
> @katomm yes please merge this and remove adapools.

 When I wrote "we" I meant "us all" :)) Possible for you to just remove the entry and delete the image?

katomm (approved):

PR #749, opened 2022-08-29
Add new filter to check latest showcases and builder tools 

fill-the-fill:
I'm currently working on a new feature for users to filter showcases and builder tools by latest projects. I would like to hear your opinion on it. 

Currently I'm displaying LAST 10 projects added in the filter. What do you think?

filter

### DIFF
--- a/src/components/showcase/ShowcaseFilterToggle/index.js
+++ b/src/components/showcase/ShowcaseFilterToggle/index.js
@@ -25,7 +25,7 @@ export default function ShowcaseFilterToggle() {
   const id = "showcase_filter_toggle";
   const location = useLocation();
   const history = useHistory();
-  const [operator, setOperator] = useState(true);
+  const [operator, setOperator] = useState(false);
 
   useEffect(() => {
     setOperator(readOperator(location.search) === "OR");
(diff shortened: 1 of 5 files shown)


rdlrt:
It's awesome to add that in, only thing I'd wonder us are you intending it to be default filter?

fill-the-fill:
> It's awesome to add that in, only thing I'd wonder us are you intending it to be default filter?

By default it's showing all of the projects

rdlrt:
Cool, only thing I could think of is name (Last va Latest), but either works (esp. as Latest is 6 letters, mught be less consistent visibly)

fill-the-fill:
> Cool, only thing I could think of is name (Last va Latest), but either works (esp. as Latest is 6 letters, mught be less consistent visibly)
A bit off imo, unless we come up with better design 

Screenshot 2022-08-29 at 11 38 56

rdlrt:
Agree, let's leave it as you have it in PR

rdlrt (approved):

rphair (approved):
nice work 🤓

os11k (approved):

thenic95:
Awesome @fill-the-fill 👍

PR #753, opened 2022-09-01
Add a new builder tool tag - PureScript

fill-the-fill:
Current staging is failing to build, since one of the builder tools had a non existing tag. I've added PureScript as another filter in Builder Tools.

I would like to hear your opinion if we should add it, or remove the tag completely?

### DIFF
--- a/src/data/builder-tools.js
+++ b/src/data/builder-tools.js
@@ -175,6 +175,14 @@ export const Tags = {
     color: '#7e6a4c',
   },
 
+  // Purescript
+  purescript: {
+    label: "Purescript",
+    description:
+      "PureScript language",
+    icon: null,
+    color: '#0F9D58',
+  },
 };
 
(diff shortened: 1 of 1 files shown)


rphair (approved):
my apologies, I couldn't test as usual since packing today & running to airport tomorrow for intercontinental move... I hadn't seen the `purescript` tag before & assumed the default behaviour would be to ignore it 😖

I think PureScript is still part of Cardano's proposed development path as documented here: https://why.cardano.org/en/science-and-engineering/why-haskell/

There is no mention of it since 2021 on the Cardano Forum but it looks like it could still be used as part of a niche community.  I'm not familiar with the language or use cases but I can imagine the PureScript tag would be of great interest to that community looking to share ideas and tools.

rdlrt (approved):

katomm (approved):
👍

PR #762, opened 2022-09-07
Futurbnk - Cardano Token / NFT Minter Tool

learncardano:
## Showcase addition

* Name: **Futurbnk**
* Description: Futurbnk provides CNFT creators and game developers a very simple to use online tool to mint fungible and non-fungible tokens.
* Link:   [https://futurbnk.io/](https://futurbnk.io/)
* Source: 
* Tags:
  * `minting`
  * `nftsupport`


rdlrt:
I am really sorry to add a negative review, but in it's current state - this is not a wallet I would recommend *ANYONE* to use for crypto funds :
- It is a shared-custodial (i.e. user mnemonics may be shared, but encrypted private keys are always shared with remote server) wallet, but does not mention so on the website
- Once user creates a wallet, it presents an option to backup wallet keys, but even if "backup" is not checked - it ends up sharing keys to remote server using form data
- It uses a component that has proved to not be scalable for large number of users (IMO wasnt designed to do so).

I believe all of this is a pattern that emerges due to use of cardano-wallet as a component instead of cryptographic libraries that should not require a user to trust a remote server. Even if your intent might be the aiming for best user experience, it exposes a non tech-savvy user to a risk.

Again, apologies for not going ahead with approval on this one, hopefully the site can be worked to become trustless

learncardano:
> I am really sorry to add a negative review, but in it's current state - this is not a wallet I would recommend _ANYONE_ to use for crypto funds :
> 
> * It is a shared-custodial (i.e. user mnemonics may be shared, but encrypted private keys are always shared with remote server) wallet, but does not mention so on the website
> * Once user creates a wallet, it presents an option to backup wallet keys, but even if "backup" is not checked - it ends up sharing keys to remote server using form data
> * It uses a component that has proved to not be scalable for large number of users (IMO wasnt designed to do so).
> 
> I believe all of this is a pattern that emerges due to use of cardano-wallet as a component instead of cryptographic libraries that should not require a user to trust a remote server. Even if your intent might be the aiming for best user experience, it exposes a non tech-savvy user to a risk.
> 
> Again, apologies for not going ahead with approval on this one, hopefully the site can be worked to become trustless

Thank you for your review. Your feedback is truly valuable for us. We will do our best and strive to improve on our project based on your review as we go along. Although our aim is not really to become a custodial wallet for storing large amount of assets (there are much better and more secured wallets out there), but the wallet feature is only an added tool for keeping the assets (fungible and non-fungible) that our users mint on our platform, which they can send out later to a more secured wallet they have. As mentioned in our security primer page, we don't recommend users to keep their assets in our wallet. But nonetheless, this is just a trial submission hoping we can pass your requirements, as there seems to be links in the showcase page to non-existent wallets or url.

rdlrt:
> as there seems to be links in the showcase page to non-existent wallets or url.

oh - could you expand on which wallet links are those? I tried to look at [all wallets](https://developers.cardano.org/showcase/?tags=wallet) but could not find one that's non-existing, we should rectify if any projects have stopped since

learncardano:
> > as there seems to be links in the showcase page to non-existent wallets or url.
> 
> oh - could you expand on which wallet links are those? I tried to look at [all wallets](https://developers.cardano.org/showcase/?tags=wallet) but could not find one that's non-existing, we should rectify if any projects have stopped since

The GameChanger link is dead. Maybe they just changed their url, I don't know. It's the same reason we built Futurbnk, we were aiming for newbies, students and NFT creator wannabies who are not very, as you mentioned, "tech-savvy", and provide them a very simple to use tool. Futurbnk was not meant to compete with any wallet, we just want to promote our very easy to use NFT minter tool as a business. ;-)

learncardano:
Can you at least add Futurbnk under Minting tag? We will just change our web site contents and change our image as an NFT minting tool instead of a wallet.

learncardano:
Or even Educational tag as I will be posting series of tutorials too and know-hows about Cardano assets, NFTs, etc. on our web site which I have been doing on my [youtube channel](https://www.youtube.com/channel/UCLuVvlVKKJ5o2ly3QDIU81w)

rdlrt:
It's prolly best for you to have a product that you're aiming for implemented , and then create a PR - I don't think any of us can commit to future integration on the basis of current state, as one of the prereqs to be listed on  is working [live] product that's being used by users, and the current website is predominantly presenting a light wallet (unless there are different opinions from others) 👍🏻

learncardano:
You're right I guess we made a mistake in highlighting the wallet feature instead of the minter tool that we are promoting. I guess we will revise our site contents and rebrand our web site as an NFT minting tool rather than custodial wallet which is not our objective. Thank you for all the info.

rdlrt:
Thank you for your immediate actions, but for me, as regards security - the base issue remains

Any wallet that holds user funds, and exposes it in form data (as below), regardless of how it is / isn't used on server end - onboards risk to keys (be it at MITM, proxy level or otherwise). It was a different thing when ecosystem did not have the capabilities to interact with user wallets without them giving up any control. Even in this case, a NFT created by someone is subject to loss of control unless immediately transferred off their own wallets, there should be NO reason to have something like below in POST requests to remote servers.

IMG_20220907_182351_325


But I will defer to others to comment 🙂

learncardano:
Got you on that one, it's actually already in our minds, I guess to avoid http posting the plain text (mnemonic), which is your primary security concern, then we may have to let go of the capability to allow our users view on screen and write down their mnemonic. And secondly, we were thinking of simply letting go of the functionality to restore existing wallets. Therefore all accounts created on our web site and brand new wallets and not recovery phrase ever be shown on screen.

learncardano:
We have removed the wallet restoration feature upon sign up so no posting of mnemonic occurs. But for now we kept the viewing of the fresh new mnemonic on screen to allow the user to write it down should the user decides not to let us backup the mnemonic. While we figure out more security measures to apply. Thank you for all your feedback.

katomm:
Amazing catch @rdlrt. Closing this due to the above-mentioned concerns.

PR #767, opened 2022-09-09
Rework labels & filters on builder-tools

KtorZ:
Notable changes:

  #### Rework labels in two sections: languages vs domains

  This makes it easier to navigate; one usually wants to select some technlogy they're familiar with, and then, some particular domain of application or problems they're trying need a tool for.
  There's probably some additional rework to do with the domains, but I've added a few to try to get at least every tool with a language and a domain that made sense. Consequently, I've also
  removed the 'library' label as it is implied by the presence of a language label, whereas APIs and services have new labels "http" or "websocket" depending on the case.

  #### Remove 'getting started' from labels

  The semantic of this label was a bit confusing IMO. One would thing that only "beginner friendly" tools would have this label but not, it only identifies
  tools / libraries that have a "getting started" page -- though the definition of a "getting started" page isn't quite clear. Most projects have README which would
  qualify as 'getting started'. I've left the "Getting Started" button, but removed it from the label to avoid cluttering the label list needlessly.

  #### Remove 'favorite' from filters

There's already a whole section with all the favorites right below. So the filtering already happens anyway.

  #### Remove 'libada-go' from the favorite

  This library is unmaintained and hasn't moved in the past 13 months. I don't think we should point it to people as a "favorite".

---

Screenshot from 2022-09-09 13-16-32

### DIFF
--- a/src/data/builder-tools.js
+++ b/src/data/builder-tools.js
@@ -28,14 +28,6 @@ export const Tags = {
     color: '#e9669e',
   },
 
-  // API
-  api: {
-    label: "API",
-    description: "Cardano API.",
-    icon: null,
-    color: '#39ca30',
-  },
-
   // Chain Index
   chainindex: {
     label: "Chain Index",
@@ -50,20 +42,12 @@ export const Tags = {
     label: "CLI",
     description: "Command-line interface tools",
     icon: null,
-    color: '#fce300',
-  },
-
-  // For builder tools with a get started tag, a link to the get started page is required.
-  getstarted: {
-    label: "Get Started",
-    description: "This builder tool has a get started page in the developer portal.",
-    icon: null,
-    color: '#dfd545',
+    color: '#921f32',
   },
 
   // Golang
   golang: {
-    label: "golang",
+    label: "Go",
     description:
       "Go language",
     icon: null,
@@ -86,7 +70,7 @@ export const Tags = {
       "Java language",
     icon: null,
     color: '#d5232d',
-    },
+  },
 
   // JavaScript
   javascript: {
@@ -94,16 +78,7 @@ export const Tags = {
     description:
       "JavaScript language",
     icon: null,
-    color: '#921f32',
-    },
-
-  // Library
-  library: {
-    label: "Library",
-    description:
-      "Cardano library.",
-    icon: null,
-    color: '#a44fb7',
+    color: '#fce300',
   },
 
    // Marlowe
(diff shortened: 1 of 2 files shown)


fill-the-fill:
I think we shouldn't be removing `favourite` filter. It makes it easier to filter when combining with other filters. For example: If someone wants to filter only favourite APIS, it makes things much easier.

KtorZ:
@fill-the-fill there's already a list of favorite elements filtered below under the "Favorite" section. The "favorite" filter AND the favorite section are redundant with one another. Keeping both doesn't make much sense to me; keeping only the filter however (and removing the section) would make sense!

katomm:
If we want to remove the getstarted labels (which are indeed only marking if one has a get started section in the portal, we need to change also the build script. It enforces a get started label for get started page as far as I remember. (Means this PR would fail with yarn build)

KtorZ:
@katomm I did change the build script already --> https://github.com/cardano-foundation/developer-portal/pull/767/files#diff-53a90cc1cf2c5306e1f7c14e7fdc2e148e8673f81d09335596b257bc2e65dc3aL758-L770

katomm:
@KtorZ Sorry, I missed that, thought it was in a different file. Any reason for also removing the check to put operator tools below /docs/operate-a-stake-pool?

katomm (approved):

KtorZ:
Whoops. No. Wil revert.

rphair (approved):
@KtorZ @katomm I'm just looking at a built preview of this now after following the discussion for a few days.  It really is a huge improvement and it looks like the keywords for all projects have been corrected in some cases in addition to being extended with the new tag vocabulary.

I'm also happy to verify it builds perfectly with `staging` as it exists at this time 🤓

KtorZ:
>  @rphair: it looks like the keywords for all projects have been corrected in some cases in addition to being extended with the new tag vocabulary.

Indeed, I've made some rather opinionated choices which I think are better. Up for discussion obviously and I imagine this to be an ongoing effort. Note that there are at least two similar efforts going on within the CF right now (on different fronts, one being CIPs themselves) to define a spectrum of "business domains" to better categorize tools or even CIPs. Ideally, I wish for all these efforts to converge into one terminology that can reuse across the boards and with which people can get familiar with. 

So, this PR is a first stab at it. Not perfect yet but I hope a slightly better increment to the current status quo.

rdlrt (approved):

PR #769, opened 2022-09-10
Fix cardano-node build instructions on recent MacOS M1

KtorZ:
On a brand-new MacOS w/ M1 (not sure if the problem really is the ARM arch here, or simply that homebrew has changed default installation paths of openssl libs & includes); exact same issue also hit by at least one other MacOS M1 user. 

Moreover It seems that on MacOS, the [`HsOpenSSL`](https://github.com/haskell-cryptography/HsOpenSSL/blob/master/HsOpenSSL.cabal#L90-L96) haskell package also require an extra flag (`-homebrew-openssl` or `-macport-openssl` depending on the mean of installation) without what it will simply not find OpenSSL. 

Screenshot from 2022-09-10 08-59-32


katomm (approved):
👍

rphair (approved):
:apple: :grinning:

fill-the-fill (approved):

PR #771, opened 2022-09-12
Hint: remove leading whitespace from all JSON lines before pool metadata hash

ubani:
Added a debugging hint regarding metadata hash.

### DIFF
--- a/docs/stake-pool-course/handbook/register-stake-pool-metadata.md
+++ b/docs/stake-pool-course/handbook/register-stake-pool-metadata.md
@@ -48,7 +48,7 @@ Registering your stake pool requires:
 
 Store the file on your web server. For example [https://teststakepool.com/poolMetadata.json](https://git.io/JJWdJ). You must store your pool metadata on reliable server, in case if metadata files will not be available then your pool might disappear from wallets and potential delegators will not be able to find it.
 
-**HINT:** Remove preceeding whitespaces from your JSON file. While some tools are more generous when it comes to comparing the metadata hash, others will only accept a correct match. In case of doubt compare your hash with pooltool.io(https://pooltool.io) or with the transaction output on https://cardanoscan.io(https://cardanoscan.io)   
+**HINT:** Remove preceding whitespace from your JSON file. While some tools are more generous when it comes to comparing the metadata hash, others will only accept a correct match. In case of doubt compare your hash with [pooltool.io](https://pooltool.io) or with the transaction output on [cardanoscan.io](https://cardanoscan.io).
 
 #### Get the hash of your metadata file uploaded to server:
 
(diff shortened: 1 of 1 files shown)


docs/stake-pool-course/handbook/register-stake-pool-metadata.md
```diff
@@ -48,6 +48,8 @@ Registering your stake pool requires:
 
 Store the file on your web server. For example [https://teststakepool.com/poolMetadata.json](https://git.io/JJWdJ). You must store your pool metadata on reliable server, in case if metadata files will not be available then your pool might disappear from wallets and potential delegators will not be able to find it.
 
+**HINT:** Remove preceeding whitespaces from your JSON file. While some tools are more generous when it comes to comparing the metadata hash, others will only accept a correct match. In case of doubt compare your hash with pooltool.io(https://pooltool.io) or with the transaction output on https://cardanoscan.io(https://cardanoscan.io)
```
rphair, line 51:
```suggestion
**HINT:** Remove preceding whitespace from your JSON file. While some tools are more generous when it comes to comparing the metadata hash, others will only accept a correct match. In case of doubt compare your hash with [pooltool.io](https://pooltool.io) or with the transaction output on [cardanoscan.io](https://cardanoscan.io).
```

rphair (changes requested):
This is not strictly necessary and could be misleading: JSON is understood to use white space to indicate nesting by indentation, so to tell users to remove all white space would be unprecedented.

In any case the spelling and markdown formatting errors would have to be cleaned up first:

docs/stake-pool-course/handbook/register-stake-pool-metadata.md
ubani: Thx for the update

ubani:
TBH the spaces gave me quite a headache when I tried to match the hashes.

rphair (approved):

os11k:
Hi! Not sure if this is needed at all, if you check next paragraph, then there is stated how to calculate hash from file what is uploaded to remote server already. So if remote server has some extra spaces(new line for example at the end of file), then this hash will be calculated accordingly. I do understand a problem when some manuals shows how to calculate hash from local file and local file might differ from remote file, but again in our case it should not be a problem. Would be nice to figure out what exactly issue OP faced and how exactly that hint helped him fix that problem...

gitmachtl:
hash is calculated from the binary data, not from a parsed data. most of the time people are messing up the CR/LF or CR thing. by doing a copy&paste of the data. the metadata file must be uploaded/downloaded in a binary way. whitespaces or other stuff are not important. but the files must be the same on the binary (not text) level.

rphair:
@gitmachtl I agree and for this reason @ubani I think whatever "Hint" there is in this section should instead focus on the common source of error themselves, e.g.:
- using copy & paste on screen, which turns tabs into spaces (_Hint:_ don't do it that way)
- Windows vs. DOS style line termination differ between those 2 environments (_Hint:_ don't edit in both environments, and ensure an exact file copy between desktop & web server with `rsync` or equivalent)

rphair:
@ubani I'm going to close this one because I think the changes as offered won't suit the community for reasons already offered above (https://github.com/cardano-foundation/developer-portal/pull/771#issuecomment-1247858746).  I do believe the problem you see could be addressed if changed a bit according to my last comment (by alerting users to white space differences & what they can do about them, while necessarily preserving the indentation supported by the JSON standard).

**Other reviewers**: if it appears rash to be closing this without unilateral agreement (which is hard to get on some of these stale PRs), then if anyone claims the JSON white space is really a problem just please say so & I'll modify this PR myself, with the alternate wording suggested above, and submit it again for your review.

PR #772, opened 2022-09-12
Change base16 encoding mechanism

jsyme222:
I have altered the lines to base16 encode the token names using `basenc` instead of `xxd`. `xxd` is not part of `coreutils` and isn't found on some machines (the cardano-node docker container for instance).

### DIFF
--- a/docs/native-tokens/minting.md
+++ b/docs/native-tokens/minting.md
@@ -101,12 +101,12 @@ You need to adjust the path on your setup and your socket path accordingly.
 Since we've already answered all of the questions above, we will set variables on our terminal/bash to make readability a bit easier.
 We also will be using the testnet. The only difference between minting native assets in the mainnet will be that you need to substitute the network variable <i>testnet</i> with mainnet. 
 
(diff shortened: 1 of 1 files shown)


rphair:
This is an important point about the Docker container that should be addressed somehow, but I wouldn't want to strand users on other platforms: e.g. on Ubuntu `xxd` is universal because it's a `vim` dependency:
```
$ apt-cache rdepends xxd
 |pollinate
    xxd:i386
  vim-common
    xxd:i386
```
... while the `coreutils` on Ubuntu doesn't include `basenc`:
```
# apt install coreutils
coreutils is already the newest version (8.30-3ubuntu2).
# suspend
$ basenc
basenc: command not found
```

jsyme222:
> ... while the `coreutils` on Ubuntu (checked also on `22.04.1`) doesn't include `basenc`:
> 
> ```
> # apt install coreutils
> coreutils is already the newest version (8.30-3ubuntu2).
> # suspend
> $ basenc
> basenc: command not found
> ```

I didn't know that. Thank you.

PR #776, opened 2022-09-15
Add Github Changelog 

fill-the-fill:
The idea is to replace Weekly Report with a Changelog that will document all of the merged actions on Github. The file will be updated every month (instead of current every week) and with help of Docusaurus plugin will display all of it on Developer Portal.

Preview:
Screenshot 2022-09-15 at 12 06 20

Screenshot 2022-09-15 at 12 13 50


rphair (approved):
Looks nice & I feel sure the community will appreciate it 🤓
I can confirm it builds properly with the current `staging` ✔️

I'd never seen that way of displaying date ranges before (year followed by week number, then what might be a serial number usually `0`?) but I'm not bothered by it... just would like to know this method is called, if anything 🧐

katomm:
@rphair it's a proposal for versioning: <major>.<minor>.<revision> that uses <year>.<week>.<revision> that we thought might make more sense for ongoing things like the developer portal instead of doing V1.1, V1.2 etc

rdlrt (approved):

katomm (approved):

PR #774, opened 2022-09-14
Add Cardano Signer builder tool

fill-the-fill:
* Name: *Cardano Signer*
* Description: *Tool to sign data with a Cardano-Secret-Key and verify data with a Cardano-Public-Key*
* Link:  [Cardano Signer](https://github.com/gitmachtl/cardano-related-stuff/tree/master/cardano-signer)
* Source: [Cardano Related Stuff](https://github.com/gitmachtl/cardano-related-stuff)
* Tags:
  * `operatortool`

PR #777, opened 2022-09-17
cardano-signer: replace narrative thumbnail with logo

rphair:
[updating for tool version `1.4.0`] as @gitmachtl requested in https://github.com/cardano-foundation/developer-portal/pull/774#issuecomment-1249916232.

@katomm @fill-the-fill I felt this was worth rendering in great detail because the usage string indeed shows more capabilities, but (in hindsight) is there any way this full resolution image will even show in the Dev Portal front end?  As far as I can tell the user experience is only of a clipped thumbnail which, if opening the clipped image separately, shows a downscaled image 640p wide. 🧐

i.e. at what point would a text-based image like this make sense to update, if the end user can't read the text?


gitmachtl:
ok maybe only the logo somehow?

image

image

rphair:
yes, I saw the logo & thought of that possibility, but 1) the logo doesn't say anything beyond what's already in the UI title, i.e. nothing about what the tool is or does... and 2) I think @fickevics-makor was on the right track somehow in using the text as a demo image... we just need to know in what context that full resolution image might appear 🤓

katomm:
In the build process Docusaurus creates several downscaled images from the full resolution to deliver via web server later. Other than that the full resolution image just bloats the repo when checking out. (Which is, in my opinion, why we should care a little on image sizes as it makes no sense to have huge image file)

My recommendation is to go with the logo.

gitmachtl:
its the wrong thread here, but maybe someone can help me to bring the "StakePoolOperator Scripts" to the developer portal? i am so busy and never had the time to dig into the dev portal usage... but they should be on the portal, many are using them on a daily base.

url: https://github.com/gitmachtl/scripts
readme: https://github.com/gitmachtl/scripts/blob/master/cardano/mainnet/README.md

rphair:
> @rphair ah sorry, i thought the picture was for the thumbnail logo in the list, not a demo picture.

@gitmachtl you thought correctly, but originally (https://github.com/cardano-foundation/developer-portal/pull/774) a demo picture was provided as a thumbnail picture.  I assumed that @fill-the-fill had good reason for doing that, so updated the image accordingly.

Before we decide how to screenshot https://github.com/gitmachtl/scripts we should wait for this thread to be decided first: then I think the precedent is to create another PR, which I will do if you're not familiar with the dev environment.

@katomm this image is indeed about 80th percentile in size compared to the other tools, but I want to hear from @fill-the-fill first about why he chose that method before throwing that detail away.

katomm:
> but maybe someone can help me to bring the "StakePoolOperator Scripts" to the developer portal

We got you. I created an issue for that.

rphair:
It has somewhat belatedly occurred to me that this command's syntax and/or capabilities will also be changing in the future (as they have since version `1.2.0`) and we don't want to have to keep changing the featured image.  So after all that work 😭 it does indeed make more sense to use the logo, which will also match the new builder-tool item requested in https://github.com/cardano-foundation/developer-portal/issues/778.

katomm (approved):
👍

gitmachtl (approved):

rdlrt (approved):

PR #788, opened 2022-09-20
Various open graph images

katomm:
This PR has many different open graph images(they define how sites look like when being shared) added: get started, dev blogs (instead of just having a default) I was not able to create/change all but this should improve the situation a lot. 

Unfortunately, small changes had to be made to many files to do this. I also create a folder in images "og" for open graph to have these previews there. If you want to read more about the open graph standard please visit: https://ogp.me


rphair (approved):
that's an ambitious improvement 😍 and builds OK with current `staging` ✔️

I tried using this browser extension (https://chrome.google.com/webstore/detail/localhost-open-graph-chec/gcbnmkhkglonipggglncobhklaegphgn) but couldn't verify that the `og:image` values were passing through from the locally built pages to the outside page which is supposed to contain the scraped metadata.  It could well be this extension isn't working right or I don't know how to use it properly.

But the `property="og:image"` `<meta>` tags look OK visually... in the same format as before & with the new images... so I don't see why it wouldn't work after being published: though I would recommend double checking after `staging` is moved into production 😎

fill-the-fill (approved):

rdlrt (approved):

katomm:
Yes it is a little bit painful to test that on a local machine, I also noticed a few problems already and will soon add another patch on this. On top of that also good to know: all the services cache these image files very hard, and changing them afterwards usually require to use tools where you can submit a new link. Twitter, LinkedIn, Telegram etc they all have different sites and tools for this.

PR #791, opened 2022-09-21
Installing cardano-node: MacOS missing openssl

pgrange:
Thanks for the Mac installation doc. It worked perfectly (Intel) but I was missing openssl. I though it would be useful to add it in the brew packages list.

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -290,6 +290,7 @@ brew install libtool
 brew install autoconf
 brew install automake
 brew install pkg-config
+brew install openssl
 ```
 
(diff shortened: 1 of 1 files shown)


rphair (approved):
I don't have a Mac to test this on, but the package added in this PR looks like a prerequisite for the instructions in this section (and will be required for users who haven't installed the `openssl` package for some other reason): https://developers.cardano.org/docs/get-started/installing-cardano-node#you-will-need-to-run-following-commands-on-m1-those-commands-will-set-some-cabal-related-options-before-building

Builds OK with current `staging` ✔️

rdlrt (approved):

os11k (approved):

PR #792, opened 2022-09-22
Add "known problems"-section to "how to contribute?"

katomm:
This PR adds hints for some common problems that people might run into when they install the developer portal locally. I used most of the error messages in titles so that I can be found via search.


docs/portal-contribute.md
```diff
@@ -179,3 +179,29 @@ developer-portal
...
+We list here problems you may run into when running the developer portal locally.
+
+### Minimum Node.js version not met 
+**Problem:** `yarn start` throws the error `[ERROR] Minimum Node.js version not met :(`.  
+**Solution:** use the node version listed below [requirements](#requirements). If you have different node versions installed for different projects, `nvm` is a neat tool to deal with it. You can switch versions with for example `nvm use 16`.
+
+### Sidebars file at "developer-portal/sidebars.js" failed to be loaded
+**Problem:** `yarn start` throws the error `[ERROR] Sidebars file at "developer-portal/sidebars.js" failed to be loaded.`. 
+
+**Solution:** you need to run at least once `yarn build` as this pulls missing files into your folder, which is then referenced by sidebars.js.
```
rphair, line 193:
```suggestion
**Solution:** you need to run at least once `yarn build` as this pulls missing files into your folder, which is then referenced by `sidebars.js`.
```

docs/portal-contribute.md
```diff
@@ -179,3 +179,29 @@ developer-portal
...
+
+### Sidebar category Token Registry has neither any subitem nor a link
+**Problem:** `yarn start` throws the errow `[ERROR] Error: Sidebar category Token Registry has neither any subitem nor a link. This makes this item not able to link to anything.`.  
+
+**Solution:** you need to run at least once `yarn build` as this pulls missing files into your folder, which is then referenced by sidebars.js.
```
rphair, line 198:
```suggestion
**Solution:** you need to run at least once `yarn build` as this pulls missing files into your folder, which is then referenced by `sidebars.js`.
```

docs/portal-contribute.md
```diff
@@ -179,3 +179,29 @@ developer-portal
...
+
+## Other questions
+Various other questions and answers.
+
+### Should I commit the yarn.lock file?
+No, please do not commit your `yarn.lock` file it is an auto-generated file that should be handled only by `yarn`. 
+
+### Shouldn't yarn.lock be in the ignore file then?
+No, as this will have side effects with dependencies used by the project itself. (Even if you prefer npm)
```
rphair:
```suggestion
No, as this will have side effects with dependencies used by the project itself. (Even if you prefer `npm`)
```

rphair (approved):
Looks great.  Only suggestions below are to keep to the computer monospace consistently (in the paragraphs, not in the titles) 😎

I've run into some additional cases where I couldn't build the portal and/or switch between branches without doing something additional with `yarn.lock`, but I think these were all from me either committing `yarn.lock` by mistake or testing PR's in which that had happened.  Resolving these cases was either a general `git` issue (fixable though `git`'s own documentation) or literally following the command line responses from `git`... so I don't think they need to be addressed here, but I'll take a closer look the next time this happens 🤓

docs/portal-contribute.md
katomm: Good point.

docs/portal-contribute.md
katomm: Agree

fill-the-fill (approved):

os11k (approved):

PR #795, opened 2022-09-22
Open graph images patch

katomm:
This PR will fix issues introduced with PR #788 and update more pages to use a specific open graph image instead of the default one, that only shows the start page of the developer portal.

Screenshot 2022-09-22 at 13 08 52

### DIFF
--- a/blog/2021-01-07-january.md
+++ b/blog/2021-01-07-january.md
@@ -7,7 +7,7 @@ author_url: https://github.com/cardano-foundation
 author_image_url: https://avatars.githubusercontent.com/u/37078161?s=200&v=4
 tags: [pooltools]
 description: Cardano Developer Spotlight January 2021
-image: https://developers.cardano.org/img/og-blog-adatools.png
+image: https://developers.cardano.org/img/og/og-blog-adatools.png
 ---
 
 Welcome to the January Developer Spotlight 2021. This month, we will focus on pool tools. Read on to learn about the **ADATools.io** and the **Pool.pm**.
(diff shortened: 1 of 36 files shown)


fill-the-fill (approved):

rphair (approved):
great 😍 ... @katomm does the previewer screenshotted in the OP work on locally served web pages? (re: https://github.com/cardano-foundation/developer-portal/pull/788#issuecomment-1254652544)

Builds OK with current `staging` ✔️

katomm:
@rphair it depends, ofc not with tools like https://www.opengraph.xyz because they can not access the site. And other local tools will run into issues trying to find the new open graph image because it looks on developers.cardano.org domain even if we put in the link relative and not absolute. It would work locally if the image is already in the img/og/ folder, then it just fetches the image from the cardano.org domain.

It is some hassle but we should go through it because the sharing experience for users on socials is so much better.

rphair:
@katomm from your comments in both PR's, and the official looking screenshot you have above, I'm guessing you're using [ngrok](https://ngrok.com) or equivalent to set up a web accessible version of the localhost site?  Especially if that equivalent will map `developers.cardano.org` to the domain name of the copied site, according to your last comment?

rdlrt (approved):

katomm:
> and the official looking screenshot you have above, I'm guessing you're using [ngrok](https://ngrok.com) or equivalent to set up a web accessible version of the localhost site?

Not at all. That is a real screenshot. Might be a little bit misleading because this change was already done in PR #788, it was not meant to visualize the changes in this PR :)) But ofc they are similar.

PR #798, opened 2022-09-25
Updated script for getting pool statistic

os11k:

Updated script for getting pool statistic, now we will use cexplorer api instead of deprecated and not updated adapools api. Updated script creates sames metrics, so user still can use SNSKY dashboard and most important, existing user can update old script with one from here and old dashboard will work out of the box. There are some metrics is missing, but in case if SNSKY dashboard is being used, then it is only one metric(adapools_pledged), so I would say it is not perfect, but more then ok, IMHO.


rphair (approved):
it works great 🤠 and PR builds OK with current `staging` ✔️

rdlrt (approved):

fill-the-fill (approved):

PR #826, opened 2022-10-14
Revamp stake pool course | Adding Pre-requisites and Hardware Requirements

sanskys:
#### Issue

Fixes #737 

adding pre-requisites and hardware-requirements

PR #829, opened 2022-10-14
Delete yarn.lock from revamp-stake-pool-course

rphair:
Somehow `yarn.lock` got checked in to this branch (it is NOT in the `staging` branch of the main repository)... which prevents contributors from switching to a local testing copy from any other branch which builds `yarn.lock` as we are always supposed to do (right?).  If there was something special about `yarn.lock` in [revamp-stake-pool-course](https://github.com/cardano-foundation/developer-portal/tree/revamp-stake-pool-course) that we need to keep, please say so 😎


rphair:
OK, it looks like this has been taken care of with https://github.com/cardano-foundation/developer-portal/pull/826 ... didn't occur to me until now all I had to do is push this commit into some other contributor's branch & I could have approved it myself 🤣

PR #832, opened 2022-10-17
Add yarn build workflow

oversize:
## Add github workflow for yarn build 

This change introduces a github workflow that tries to build the documentation with
yarn. It does so by first running `yarn install`  followed by `yarn build`.
See .github/workflows/yarn-build.yml. 

I have added the yarn.lock file to the repository, to have deterministic dependency
installations. See yarn docs.

The github action itself is implemented in .github/actions/yarn-build. I wanted to use a 
provided one but had to add a special fix for a git feature recenetly added, which caused
alot of warnings in the build log. I have added a README in the action with more details.

Currently the yarn build will run on every push and every PR. This might be something to discuss.


katomm (approved):
Thank you for this Manuel.

rphair (approved):
:nerd_face: :pray:

oversize:
@rphair yes, the only time the yarn.lock should change is through the yarn cli (adding, deleting dependencies etc).

However, a new lock file will not be generated if you just issue a yarn install though. If there is a lock file present, yarn install will use the dependency tree from that file. I am not a yarn use but that would be my assumption based on how pipenv works with lockfiles.

rphair:
@oversize we periodically had cases like recently resolved in https://github.com/cardano-foundation/developer-portal/pull/829 and also in individual PR's when a `yarn.lock` produced in a branch would need to be deleted before other users could pull & test that branch properly (e.g. it would prevent leaving the branch if the `yarn.lock` in it wasn't committed or stashed, and prevent entering the branch if a `yarn.lock` generated elsewhere would be overwritten).

So it's not a problem with the new dependencies that are generated but the ability to switch between different branches.  I just have no idea how this will affect testing if different local branches, all with concurrent PR's, end up committing slightly different versions of `yarn.lock`.  I don't think that effects the usefulness of this new build facility... it's just something else for us to be aware of in testing the next few PR's. 😎

oversize:
Yes, as long as there are branches (and ppl are working on them) that dont have a yarn.lock file, changing from these branches will be a problem. However, it is my understanding that the yarn.lock file should always be commited to every branch anyhow, so, it might as well be added to each then. 

The yarn.lock is always "regeneratable" from the current package.json. You can always delete and regenerate it to "update" all dependencies and persist the new tree within that lock file. But "dependency upgrades" should only happen deliberately. In theory no one should just generate a new  lock file without actually testing all the upgraded dependencies.

rdlrt (approved):

rphair:
@oversize I was just checking to see how the local build process would get along with this, and it does indeed seem like a `yarn install` as well as the `build` will also be necessary whenever testing PR's in a local branch:
```
$ git fetch upstream pull/832/head:verify-832
 * [new ref]         refs/pull/832/head -> verify-832
$ git checkout verify-832
error: The following untracked working tree files would be overwritten by checkout:
	yarn.lock
Please move or remove them before you switch branches.
Aborting
$ rm yarn.lock
$ git checkout verify-832
$ yarn build
Internal Error: devportal@workspace:.: This package doesn't seem to be present in your lockfile; run "yarn install" to update the lockfile
...
$ yarn install
... etc ...
```
I'm OK to go ahead with this regardless, especially because if this works as I hope it will after going live, there will be less (or perhaps no more) need to do local builds.  That would be great especially since I'm working in an Asian country which has lower tier connectivity to node CDN (as well as GitHub) so the `yarn install` can be expensive & time consuming during times when international connectivity is bad.  But I'm looking forward to seeing how this goes in any case & running with it for a while 🤓

rphair:
questions on next steps:

1. @katomm @rdlrt @fill-the-fill how will we know when these commits (or the commits of any PR) go from `staging` to the production GitHub workflow?
2. @oversize how will we access the site that gets built as part of the new workflow?

rdlrt:
1. We can deploy a branch to github pages (either using [native deployment](https://docusaurus.io/docs/deployment#deploying-to-github-pages) or using a branch head as static dir browser, and each folder for currently open branches, in either case - we'd need to add and adapt a workflow, add `.nojekyll` file and then set deployment branch to the branch elected for gh-pages. When doing a push to master, you can also have same action accept a different branch or folder as target

2. I think currently it's just a test build process, to make it a site we need to go back to point 1 aboce

oversize:
Yes, currently there is no way to access what has been build. However as described by @rdlrt we easily could extend this. I would be happy to try to implement something for either one. To not explode this PRs discussion though, i could open a new issue this?

PR #834, opened 2022-10-21
Remove un-reusable functions from reusable.ts script file

fill-the-fill:
I've made some changes in `reusable.ts` and remove some of the functions that I believe shouldn't be reused. There is a lot of work to do regarding scripts to make them better, faster and cleaner, but this just a small modification for now not to confuse people when checking `reusable.ts` file.


rphair (approved):
I'm not the one to assess the code changes, but I can confirm the site builds locally & work as usual, auto-updates when files are modified, etc. ✔️

katomm (approved):
LGTM

os11k:
I'm not confident enough to approve it or not, but it does build

PR #841, opened 2022-10-26
Add STAMPD showcase

fill-the-fill:
* Name: *STAMPD*
* Description: *Use the public blockchains to timestamp your files with indelible proof and mint with linked NFC physical tags for embedment in physical objects.*
* Link:   [STAMPD](https://stampd.io/)
* Source: *null*
* Tags:
  * `minting`
  * `metadata`

stampd


rphair (approved):
Just built OK & double-checked that the writing on the landing page is still visible even when reduced to thumbnail size (it's close, but legible) ✔️

Interesting to see that the only two current projects (including this one) with both `minting` and `metadata` tags have completely unique qualities beyond these basic ones 🤓

os11k (approved):

katomm (approved):
👍

PR #842, opened 2022-10-26
Modify image for 'Cardano connect with wallet' builder tool

fill-the-fill:
I believe this image is more appropriate for the builder tool.

cardano-connect-with-wallet


fill-the-fill:
> There's more detail here that can be discerned in the thumbnail, but if people open the image separately (so it displays reduced to 640p wide) all the elements of the screen are readable. And although the aspect ratio is not 2:1 like the thumbnail image cells, the top of the image is aligned to the top of the cell... so it shows the most relevant parts of the tool UI. ✔️

I tried to include as many details as possible, this is how it looks on the website.

Screenshot 2022-10-26 at 16 34 06

os11k (approved):

katomm (approved):

Issue #835, opened 2022-10-22
Missing CIP36


gitmachtl:
CIP36 - the new standard for voting/governance/catalyst registration/delegation/deregistration - is missing in the [CIP collection](https://developers.cardano.org/docs/governance/) on the dev portal

fill-the-fill:
@gitmachtl On it, found the problem, actually we are missing more than CIP36, will be opening a PR very soon

fill-the-fill:
#848 has been merged into staging

PR #848, opened 2022-10-31
Fix displaying missing CIPs 

fill-the-fill:
As mentioned in #835, script wasn't displaying all of the CIPs, I've made some modifications.

Once again this is a temporary solution, script isn't in a good shape, I'm going to completely modify the whole thing probably this or next week.


rphair (approved):
thanks @fill-the-fill ... I've tested it locally & it seems to pick up the lot ✔️

FYI there are some more header changes on the way which will effect statuses & categorisation; in fact a new field being introduced which indicates part of the ecosystem (I see that these tags in the vocabulary are made into links).  Also some fields will allow multiple lines (author & discussion link fields, which can be bulleted lists), which could also affect the parsing: https://github.com/cardano-foundation/CIPs/pull/366

rdlrt (approved):

katomm (approved):
LGTM

PR #849, opened 2022-11-04
Revamp stake pool course | Adding Cardano Node Installation

### LINKED ISSUE #737: Revamp Stake Pool Section: new structure proposal
Because of a better overview in the Stake Pool Section, I would suggest the following new outline of the menu structure for creating and maintaining a Stake Pool in Testnet.

It is a proposal that everything listed in the current Stake Pool Operator section is replaced by a new and clearer structure.

### Overview
- [x] Get started 
- [x] Stake Pool Networking (understanding the relay and BP topology)
- [x] Pre-requisites
- [x] Hardware Requirements 
### Installation Guide (Node)
- [ ] Logging and Security (ssh key,2FA)
- [ ] Server Setup (optimize performance, chrony, updates)
- [ ] Network Setup (if needed?)
- [x] Cardano Node Installation process (Environment Setup)
### Cardano Relay Configuraion
- [x] Cardano Relay Configuration (Startup scripts, Systemd)
- [ ] Downloading Cardano Blockchain
- [x] Launching Cardano Relay Node
- [x] Monitoring the Node (Prometheus, Grafana, gLiveView)
### Cardano Block Producer Configuration
- [x] Generating wallet keys (Faucet for tADA)
- [x] Generating Cardano Block producer Keys (StakePool certificate generation)
- [x] Registering a Pool (JSON Metadata)
- [x] Launching Cardano Node (Startup scripts, Systemd)
### Stake Pool Maintenance    
- [ ] Rotate KES
- [ ] Update Registration
- [ ] Cardano Node Upgrades
### Stake Pool Operator Tools and Links (which one to choose?)
- [ ] Guild Tools (CNTools, CNCLI, etc.)
- [ ] Topology Updater
- [ ] SPO Channels

This is the first draft of this idea, please let me know what you think.

Thanks, @sanskys  for your issue #641  I hope my proposal fits your ideas.

sanskys:
Issue

Fixes #737 

Adding Cardano Node Installation

### DIFF
--- a/docs/stake-pool-guide/node-installation-process.md
+++ b/docs/stake-pool-guide/node-installation-process.md
@@ -1,0 +1,239 @@
+---
+id: node-installation-process
+title: Cardano Node Installation process (Environment Setup)
+sidebar_label: Cardano Node Installation process
+description: Cardano Node Installation process (Environment Setup)
+image: ../img/og-developer-portal.png
+---
+The Cardano node is the core component that underpins the Cardano network. Ultimately, a blockchain network is just a collection of interconnected nodes, all working together to validate transactions and blocks by means of consensus. The definition of consensus for any given network varies, but for the Cardano network it’s defined by the [Ouroboros](stake-pool-basics#understanding-consensus) protocol.
+
+In this section, we will walk you through the process of downloading, compiling, and installing `cardano-node` and `cardano-cli` into your **Linux-based** operating system.
+
+## Installing Operating System dependencies
+
+To download the source code and build it, you need the following packages and tools on your Linux system:
+
+* the version control system `git`,
+* the `gcc` C-compiler,
+* C++ support for `gcc`,
+* developer libraries for the arbitrary precision library `gmp`,
+* developer libraries for the compression library `zlib`,
+* developer libraries for `systemd`,
+* developer libraries for `ncurses`,
+* `ncurses` compatibility libraries,
+* the Haskell build tool `cabal`,
+* the GHC Haskell compiler (version `8.10.7` or above).
+
+In Redhat, Fedora, and Centos:
+```bash
+sudo yum update -y
+sudo yum install git gcc gcc-c++ tmux gmp-devel make tar xz wget zlib-devel libtool autoconf jq -y
+sudo yum install systemd-devel ncurses-devel ncurses-compat-libs -y
+```
+
+For Debian/Ubuntu, use the following instead:
+```bash
+sudo apt-get update -y && sudo apt-get upgrade -y
+sudo apt-get install automake build-essential curl pkg-config libffi-dev libgmp-dev libssl-dev libtinfo-dev libsystemd-dev zlib1g-dev make g++ tmux git jq wget libncursesw5 libtool autoconf -y
+```
+If you are using a different flavor of Linux, you will need to use the correct package manager for your platform instead of `yum` or `apt-get`, and the names of the packages you need to install might differ.
+
+## Installing GHC and Cabal
+
+The fastest way to install **GHC** (Glasgow Haskell Compiler) and **Cabal** (Common Architecture for Building Applications and Libraries) is to use [ghcup](https://www.haskell.org/ghcup).
+
+Use the following command to install `ghcup`
+```bash
+curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+```
+Please follow the instructions and provide the necessary input to the installer.
+
+`Do you want ghcup to automatically add the required PATH variable to "/home/ubuntu/.bashrc"?` - (P or enter)
+
+`Do you want to install haskell-language-server (HLS)?` - (N or enter)
+
+`Do you want to enable better integration of stack with GHCup?` - (N or enter)
+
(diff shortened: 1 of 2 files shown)


rdlrt:
> The video at the end is not up to date. Should we still keep it in?

Videos in general will always have that deficiency (even if it can be edited later, it can become clumsy) - I'd vote to remove them completely.

docs/stake-pool-guide/node-installation-process.md
```diff
@@ -0,0 +1,239 @@
...
+id: node-installation-process
+title: Cardano Node Installation process (Environment Setup)
+sidebar_label: Cardano Node Installation process
+description: Cardano Node Installation process (Environment Setup)
+image: ../img/og-developer-portal.png
+---
+The Cardano node is the core component that underpins the Cardano network. Ultimately, a blockchain network is just a collection of interconnected nodes, all working together to validate transactions and blocks by means of consensus. The definition of consensus for any given network varies, but for the Cardano network it’s defined by the [Ouroboros](stake-pool-basics#understanding-consensus) protocol.
+
+In this section, we will walk you through the process of downloading, compiling, and installing `cardano-node` and `cardano-cli` into your **Linux-based** operating system.
+
```
rphair, line 11: A link & reference needs to be added here, to the most basic (and well maintained) source file for the installation procedure: https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/install.md

As commented above regarding the videos, material can go out of date any time the tutorial stops being monitored.  Eventually that will happen this written material: in fact, very quickly after versions of GHC, Cabal, `cardano-node`, libraries, and their package dependencies change.  So since this material is mostly a re-hash of the official installation procedure, they have to be directed there for help & comparison when necessary.

docs/stake-pool-guide/node-installation-process.md
```diff
@@ -0,0 +1,239 @@
...
+For Debian/Ubuntu, use the following instead:
+```bash
+sudo apt-get update -y && sudo apt-get upgrade -y
+sudo apt-get install automake build-essential curl pkg-config libffi-dev libgmp-dev libssl-dev libtinfo-dev libsystemd-dev zlib1g-dev make g++ tmux git jq wget libncursesw5 libtool autoconf -y
+```
+If you are using a different flavor of Linux, you will need to use the correct package manager for your platform instead of `yum` or `apt-get`, and the names of the packages you need to install might differ.
+
+## Installing GHC and Cabal
+
+The fastest way to install **GHC** (Glasgow Haskell Compiler) and **Cabal** (Common Architecture for Building Applications and Libraries) is to use [ghcup](https://www.haskell.org/ghcup).
```
rphair, line 43:
```suggestion
A popular way to install **GHC** (Glasgow Haskell Compiler) and **Cabal** (Common Architecture for Building Applications and Libraries) is to use [ghcup](https://www.haskell.org/ghcup).
```
`ghcup` cannot be the "fastest" way to install GHC because the other way involves unpacking a tarball & running `make install` ... which `ghcup` with its release management also must do.

docs/stake-pool-guide/node-installation-process.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Configuring the build options
+
+We explicitly use the `ghc` version that we installed earlier. This avoids defaulting to a system version of `ghc` that might be newer or older than the one you have installed.
+
+```bash
+cabal configure --with-compiler=ghc-8.10.7
+```
+
+If you are running non x86/x64 platform (eg. ARM) please install and configure LLVM with:
```
rphair, line 196: I hadn't know this build target had become standardised, so I'm seeing it here for the first time myself: well done on including it 😎👍

rphair (commented):
Ignoring spelling & grammar mistakes, since these could be more easily fixed with a single PR when all these pages are written (and likely to be rewritten in parts) before merging into `staging` (yes I am volunteering to do the grammatical editing 😇)

docs/stake-pool-guide/node-installation-process.md
sanskys: added at the end

sanskys:
> > The video at the end is not up to date. Should we still keep it in?
> 
> Videos in general will always have that deficiency (even if it can be edited later, it can become clumsy) - I'd vote to remove them completely.

removed...we can edit and add videos once the written part is finished.

docs/stake-pool-guide/node-installation-process.md
rphair: 1. The link added at the end is *not* the official instructions for installing the node.  The link I quoted in the comment above is rigorously and responsibly maintained and is the least likely of any document to _ever_ go out of date.  So unless another editor objects, or you can offer a reason why not, I have to insist that this is the reference link included.
2. This link has to go _at the beginning of the document_, not at the end.  The reason is simple: if and when people get into problems months & years down the road, although they may still find this material useful they will need to know of an alternative source *before* they get into trouble.

I'll move & update this reference & comment in another commit (in a few minutes), so we can consider this issue tentatively resolved unless there are any objections.

rphair (approved):
All ready to go now I think; also checked build OK after my own commits 😎

PR #875, opened 2022-12-06
Add Chainport to Showcase

eserilev:
## Checklist

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Showcase addition

* Title: *Chainport*
* Description: *ChainPort is a next-gen hard-security blockchain bridge that lets you hop across EVM chains to Cardano at a click.*
* Website:   [Chainport](https://www.chainport.io/)
* Source: null
* Tags:
  * `bridge`

PR #880, opened 2022-12-09
Add Bridge filter into showcases

fill-the-fill:
## Checklist

As mentioned in #875, we don't have a Bridge filter in showcase, so I've added one. On top of that fixed some punctuation for consistency in filters description.

### DIFF
--- a/src/data/showcases.js
+++ b/src/data/showcases.js
@@ -217,7 +217,7 @@ export const Tags = {
   // Bridge 
   bridge: {
     label: "Bridge",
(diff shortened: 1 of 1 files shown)


rphair (approved):
very good: probably should add Milkomeda under separate PR, and maybe 1 or 2 more 🤔

People keep telling me this is not necessary: but at least when something new is added to the UI/UX I do a local build & I can say it builds & works well and looks good in this case ✔️

rdlrt (approved):

os11k (approved):

PR #884, opened 2022-12-13
Add Cardano Lotto to Showcase

minolinpwork:
## Checklist

## Showcase addition

* Title: Cardano Lotto
* Description: Customize and run your own lottery with a return on investment from the very first player.  A direct link is provided to promote your lottery.  Players won't have to wait for a draw and will be informed immediately if they lost / won
* Website:   https://lotto.adahouse.page
* Source: null
* Tags:
  * `game`


rphair (changes requested):
@minolinpwork I couldn't endorse this without having at least a very clear description, either on or clearly linked from the front page, about how this system works.  At this time I don't see any description there: in fact the summary you've provided here as the Showcase tagline is more information than you can find on the app landing page (identical to the parent domain `adahouse.page`)

Without a working button for `CREATE NEW LOTTERY` (presumably because I have yet done `CONNECT WALLET...`), I can't even reverse-engineer how this works.  And if people can't get any documentation on the system until after they've "subscribed" by connecting a wallet, then at best it's poorly documented and at worst could be considered exploitative.

Surely I hope to hear that there is proper documentation somewhere for this.  The "build your own lottery" premise has potential & seems more like a game of odds calculation... a mathematical exercise... than a game of chance.  But even that will require far more documentation than this tool has on the uncooperatively bare front page (e.g. choosing 7 numbers out of 99 has a "prize" of 750... for getting how many numbers right?  _All_ of them? 😜).

Another thing that would have to be documented is whether it's the web site that holds the logic for these games or a proper smart contract.  If it's just online ada payments for a web-based game of chance, I don't see that it has a place on the Developer Portal... because in that case it wouldn't be illustrating Cardano tech beyond the wallet API.

katomm (changes requested):
This would be an excellent concept for a tutorial directly in the portal, with all the source code and some text, but personally, I don't see this as a showcase.

PR #885, opened 2022-12-13
Skip creating an empty policy script

eyeinsky:
## Checklist

No need to create the empty file first.

### DIFF
--- a/docs/native-tokens/minting.md
+++ b/docs/native-tokens/minting.md
@@ -209,7 +209,7 @@ cardano-cli address key-gen \
 Use the `echo` command to create a `policy.script`:
 
 ```bash
-echo "{" >> policy/policy.script 
+echo "{" > policy/policy.script 
 echo "  \"keyHash\": \"$(cardano-cli address key-hash --payment-verification-key-file policy/policy.vkey)\"," >> policy/policy.script 
 echo "  \"type\": \"sig\"" >> policy/policy.script 
 echo "}" >> policy/policy.script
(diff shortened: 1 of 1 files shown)


rphair (dismissed):
@eyeinsky the purpose of this section is _not_ to create an empty policy script... it's actually the first step of creating the file in the next section, since it will remove any contents of the file that are already there.

However the current instructions aren't great the way they are: first because it's confusing, and second because it will always add a blank line at the beginning of the file.  I'll redraft this so the instructions are more clear overall... just stay tuned 🙏

eyeinsky:
Right, what about replacing the first `>>` with `>`.

Why I'd delete the creation of the empty file is that this code block needs not to exist at all. 90d0fbf5ee0180bdc5adfc10a3c07d8efb29fe06

rphair (approved):
This is pretty much what I was thinking, but had to add a bit more clarification for users who aren't yet able to recognise the difference between `>` and `>>`.

eyeinsky:
Cool! What I also thought for a moment was that maybe the newline at the beginning of the file was required. But then thought "that can't be true.." :p. Anyway, just trying to make it better, and these docs are pretty great when working with `cardano-cli`!

fill-the-fill (approved):

fill-the-fill:
For the future, please run `yarn build` locally before submitting the PR. That would save us some time in catching errors.

eyeinsky:
@fill-the-fill Will do!

I actually did try to run `yarn build` it, got some errors (unrelated to this PR) and stopped. But now found out that one is supposed to run `yarn` prior to that -- as a non-JS developer this wasn't obvious to me :). Anyway, have now run `yarn build` without errors.

katomm (approved):
👍

PR #896, opened 2022-12-21
Add Koios Api Python package project to builder tools

cardano-apexpool:
## Checklist

- [x] I have read the [Builder Tool Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/builder-tools.js)

## Builder Tool addition

* Title: *Koios Api*
* Description: *The Koios Api python package allows interrogating the Cardano blockchain using https://api.koios.rest/. It has pagination and retry in case of errors.* 
* Website:   [Koios Api Python package](https://github.com/cardano-apexpool/koios-api-python)
* Getstarted: [Get Started](https://github.com/cardano-apexpool/koios-api-python/blob/main/README.md)
* Tags:
  * `python`
  * `sdk`


rphair (approved):
Interesting to have this on the Portal in addition to https://github.com/cardano-foundation/developer-portal/pull/806 - at first glance it seems like the later submitted tool is better documented & may have a couple additional functions (e.g. paging).

In any case, since it has the same keywords they should be equivalently presented & perhaps the community will spontaneously provide feedback if a "better" choice develops (these tools are no competitive anyway... though we do keep an eye on usage & support).

I did remove the `getstarted:` link because its `README` content is not distinct from the project GitHub landing page.  If there is a page, now or in the future, that gives more a high-level overview then it could also be substituted here... but as presented there's no advantage to having 2 identical links.

rphair:
p.s. just looking at the built portal with both these things on it & thinking maybe we need a `koios` tag under _By language / technology_.  It seems there is room for it, and a compelling reason alone would be that we already have a Get Started page for Koios itself: https://developers.cardano.org/docs/get-started/koios/

other reviewers, what do you think about adding a `koios` tag?  I'll submit the PR for it if (all things considered including this PR) people think this would be a good idea, and add that tag to the both projects as part of that PR.

rdlrt:
> other reviewers, what do you think about adding a `koios` tag?

I'd like that as it would only work as a community collaboration, but being involved with the project - I'm obviously biased 😄 Hence, will wait for others to action

os11k (approved):

gitmachtl (approved):

gitmachtl:
> other reviewers, what do you think about adding a `koios` tag?

yup, sounds like a good idea. there is room for that imo.

PR #898, opened 2022-12-22
Add Koios tag to Builder Tools

rphair:
Pending from https://github.com/cardano-foundation/developer-portal/pull/896 and https://github.com/cardano-foundation/developer-portal/pull/806.


rphair:
p.s. see this note about why `yarn.lock` is modified in this repository: https://github.com/cardano-foundation/developer-portal/discussions/876#discussioncomment-4474103

rdlrt (approved):

gitmachtl:
so, do we tag all tools that are using koios as part of there tool now with the koios tag too? or is it only for tools theres purpose is to interact with koios as the function itself?

rphair:
@gitmachtl, the latter... "purpose is to interact with koios" as I defined it... though we could change the definition if consensus says we should.  From the tag description in the JS file (which BTW @katomm @rdlrt @fill-the-fill it would be nice to see as a "title" attribute on the tag link so it shows up as a tooltip on the web site, especially for cases like this):

> Koios and its integrations

and as I linked it on the Koios documentation page (to include "interfaces" without saying that _all_ are interfaces):

> some interfaces to Koios

In my opinion, just using Koios as part of an implementation, without the user needing to know it's there & abstracted away from the developer, isn't enough to justify tagging with `koios` & should be likewise with any similar tags.

os11k (approved):

PR #907, opened 2022-12-31
Add Cardano Studio to Showcases

bugii:
## Checklist

## Showcase addition

* Title: Cardano Studio
* Description: Create and mint NFTs on Cardano all in your browser - without trusting a third party for minting.
* Website:  https://cardano-studio.app
* Source: currently not open source
* Tags:
  * `minting`
  * `nftsupport`


rphair (commented):
I'm going to pass on this one... with the operative question being "Do we need another NFT creator besides the ones already listed here?" (https://developers.cardano.org/showcase/?tags=minting&tags=nftsupport)... but particularly because the price for NFT creators (though not yet knowing how transactions are "batched") seems compelling, I would happily hear what other reviewers think & am not suggesting it be disqualified.

Generally I would need to see documentation prior to endorsing anything.  The black wall of the landing page, with no menu or other navigation, leaves visitors literally and figuratively in the dark.  The cards for the system features would be great starting points for expanding on all these "selling points" but they're just statements with no links.

I think both some documentation and a working UI/UX should be a requirement here (besides the UX issues, on Firefox Linux the front page currently does not scroll nor present a scrollbar).

bugii:
Hi @rphair, thank you so much for reviewing and your valuable feedback.

I fixed the obvious bug with the scrollbars on Firefox and also added a navigation bar with the most important links. Also, I added a documentation and FAQ page. Of course, I plan on updating both new sites should more information be required.

rphair (commented):
@bugii thanks for making those improvements.  I read through the new pages a couple of times each and still don't see an answer for the most obvious (and probably most frequently asked) question: how & at what point you are charging the "Minimal fees = 2 ADA / transaction" (as it says on the front page) ... especially if "There is no third party you have to send ADA to" (as it says in the Documentation > Introduction).

I'll keep an eye on this & invite some other reviewers: the main question still exists of how many NFT Minting services we need to include in the Developer Portal and how this one stands in relation to the existing ones.  So I'll defer to the consensus of reviewers who have included those earlier submissions & will support your inclusion if they do. 😎

bugii:
Thanks @rphair.
So, the fee is included in the minting transaction itself. Therefore, the users mint directly using the browser extension wallets and pay the fee within the same transaction. I'll add an entry to the FAQ for this question.

bugii:
Just a short update - I made the promised minor changes.

bugii:
Just a quick check-in. Is there something I should actively work on or is it unlikely to get listed? Any feedback would be highly appreciated 🙂

rphair (approved):
I'll approve this @bugii in recognition of the work you've done so far to improve the site for general audiences.  It may be that we collectively feel another minter / NFT creator is not really needed so I am leaving that in the hands of the consensus (i.e. getting 2 more endorsements).

fill-the-fill (approved):

katomm (approved):

PR #910, opened 2023-01-03
Replace plutus-playground with demeter.run

OlofBlomqvist:
## Checklist

- [ ] I have run `yarn build` after adding my changes **without getting any errors**. 

## Updating documentation

Since the plutus playground is [no longer maintained](https://github.com/input-output-hk/plutus-apps/commit/ea11827a937fd3df58c9b590401447413b8bf762) I suggest dropping it from the docs and instead mention Demeter.run.


rphair (approved):
I haven't worked yet with Plutus, but to me this looks like the best current alternative to the [now-defunct](https://github.com/input-output-hk/plutus-apps/issues/653#issuecomment-1326453253) Plutus Playground.

os11k (approved):

rdlrt:
Agree with removal of defunct, but the replacement being suggested provides a basic sample V2 Hello World contract and VSCode workspace that provides access to shared node and CLI - not sure if it's appropriate replacement , while I appreciate what demeter brings to the table on builders section.

OlofBlomqvist:
> Agree with removal of defunct, but the replacement being suggested provides a basic sample V2 Hello World contract and VSCode workspace that provides access to shared node and CLI - not sure if it's appropriate replacement , while I appreciate what demeter brings to the table on builders section.

I think if you mean that Demeter is not providing the same feautures as the playground did, you are right. In terms of replacing the playground, Demeter just provides something rather than nothing.

That said, I believe Demeter would be worth mentioning in the docs on its own merits.

rdlrt:
> That said, I believe Demeter would be worth mentioning in the docs on its own merits.

It is currently part of [builder tools](https://developers.cardano.org/tools) (and marked as Favourite in the list). Happy for it to have it's own [get started](https://developers.cardano.org/docs/get-started/) page for it's usage under builder tools.

I do acknowledge that the gap created by lack of good tooling for Plutus does have drastic negative effects on developer onboarding experience (and much of it is an issue with closed circles that havent thought as much about giving back to community). A bit of expansion of the concerns I had are:
1. a single source of failure (hosted environment) with VSCode IDE with CLI against shared node. The samples it is addressing in current state is V2 Hello world app.
2. flow of the page : introduction of Plutus Platform follows a heading with demeter run > which ends with 'head over to demeter.run to start building', and then followed by Further tutorials from another [plutus|plutus-apps].readthedocs.io.

Instead a guide that'd focus on getting folks to build a self-sufficient environment , and then followed by reference IDEs could be better. That's just my 2 cents, others might think it's fine to have for now - happy to go with that decision too 🙂

OlofBlomqvist:
> 1. a single source of failure (hosted environment) with VSCode IDE with CLI against shared node, that is on a priced tier beyond first project. The samples it is addressing in current state is V2 Hello world script.
> 2. flow of the page : introduction of Plutus Platform follows a heading with demeter run > which ends with 'head over to demeter.run to start building', and then followed by Further tutorials from another repo [plutus|plutus-apps].readthedocs.io , which does not make use of such an environment.
> 
> Instead a guide that'd focus on getting folks to build a self-sufficient environment (contents for node/CLI are already present and can be linked), followed by reference IDEs as example environments - which could then include links to get started pages - could be better. That's just my 2 cents, others might think it's fine to have as-is for now - happy to go with that decision too 🙂

The worry I have with not mentioning any low-effort way of getting started experimenting with Plutus on this page is just that we might miss out on some otherwise would-be Cardano developers. 

Or in other words, if we can get someone to spend a few minutes tinkering with some basics they might then be more willing to put in the time and effort of installing a full local dev environment, perhaps following a guide such as the one you suggest.

How would you feel about keeping demeter here for now, moving the section to the bottom of the page and changing the ”head over there” part to something like this?:

”Demeter.run provides a free tier of their service where you can get started experimenting with Plutus”

..Or if the general opinion here is that I am wrong I will ofc just drop the demeter part from the PR :)

rphair:
> How would you feel about keeping demeter here for now, moving the section to the bottom of the page and changing the ”head over there” part to something like this?: ... ”Demeter.run provides a free tier of their service where you can get started experimenting with Plutus”

@OlofBlomqvist I'd also be open to that: if it seems to other reviewers that Demeter is currently shown with an inappropriately high rank & if demoting it a bit would resolve the apparent deadlock 🧐

rdlrt:
> How would you feel about keeping demeter here for now, moving the section to the bottom of the page and changing the ”head over there” part to something like this?

As said , it would be nice to have demeter references as long as the flow of the guide remains how-to first (same theme as compiling / installing / running node, creating keys, registrations by hand or creating NFTs by hand using local node and CLI) - and then follow it up by ready to use environments.

But in absence of that material being available, happy to go with what's easier to achieve now, as you suggest 👍🏻

rphair:
It seems very low-key now... what do you think @rdlrt?

rdlrt (approved):

PR #625, opened 2022-05-19
Updated Grafana Dashboard Tutorial

os11k:
#### Description of the change

Operate a Stake Pool => Stake Pool Course => Handbook => Grafana Dashboard Tutorial
Added small chapter to Grafana Dashboard Tutorial with steps how we can have crypto prices inside Grafana.

PR #914, opened 2023-01-05
Revamp stake pool course | Adding Grafana Tutorial

sanskys:
## Checklist

## Updating documentation or Bugfix #737 

adding Grafana Dashboard chapter


rphair (commented):
First please note I walked into this PR thinking it was an all-new page by @sanskys, so I began a full-blown editing process like for the other new course material: not realising this page has been up for a long time & in its current form since #625 (?)

Therefore I edited it substantially for some spelling (including some technical symbols) errors, English usage, and to fix punctuation & formatting as typical for computer documentation.

In any case if this is where the Grafana page is to live, I think this PR should deprecate & delete the one mounted at the old pathname (`docs/stake-pool-course/handbook/grafana-dashboard-tutorial.md` because if it's merged as-is then further editing could produce inconsistent page versions.

In general I think Section 6 needs work (though now I know it's been like that for a long time... with probably many people following the instructions successfully) because only someone going through the instructions (or who knows the setup procedure well) would be able to group these single lines into sections of what the operator is doing: either using sub-headings (as works well for the earlier sections) or using bullet points.

Now that I realise there's a history with this file I'm tagging those who've contributed to the old page.  I've actually never installed Grafana so would hope at least @os11k could review it technically (cc @rdlrt @katomm) and then if necessary I can go over the grammar & formatting again (maybe before combining this branch with `staging`).

sanskys:
The original tutorial was also written by me and chapter 6 worked well, but if it can be improved that would be great. My understanding from the beginning is that once revamp is done, the outdated stake pool course will be removed completely.

rphair:
thanks @sanskys for history and @os11k for endorsement.  Have deleted redundant page from the working branch to avoid confusion & focus any further editing on the ultimate target.  Trying to check local build but Internet connection is really bad here now; will merge when I confirm a clean build. 🐢

PR #918, opened 2023-01-08
Revamp stake pool course | Adding Grafana Security Guide

Kirael12:
## Checklist

## Updating documentation or Bugfix https://github.com/cardano-foundation/developer-portal/issues/737

Adding a new page in the Cardano Stake Pool Guide, next to Grafana Dashboard Tutorial, about how to improve Grafana Security, with SSL Reverse Proxy, and 2FA GOAuth.


rphair (commented):
thanks @Kirael12 - as I suggested before for this material (https://github.com/cardano-foundation/developer-portal/issues/737#issuecomment-1351761606) I can see from the content that it's not part of the strictly recommended setup.  Technically neither is Grafana (as you also said before) but the material here is even more of a divergence from what most SPOs need: especially if Nginx is a requirement vs. the more common Apache server.

@rdlrt I believe this is the first submitted candidate page for the special-topic "Deployment Scenarios" section (https://github.com/cardano-foundation/developer-portal/issues/737#issuecomment-1276895274) so I would propose editing this PR so it's placed under a new heading **Deployment Scenarios** on the navigation sidebar.

Kirael12:
Hello @rphair 
Thank you for your comment :) To be honest, it's my first time using Github so i was not that confortable with creating a new "section" (Deployment Scenarios). But that sounds indeed a good idea. Or maybe a "Security" section as well ?

rphair:
OK @Kirael12 since you're not comfortable with the platform & because it's a new documentation section I'll do the commit to change the file's location to a new & previously suggested _Deployment Scenarios_ section.  Stay tuned for commit...

rphair:
@Kirael12 it looks good in the new section but only in context did I realise the typographical problem of CAPITALISING ALL THE HEADER TITLES (always strange to see this in document source rather than applied via CSS).  Please keep editors in mind (as you would for all manuscripts) & don't do this again 🙏

In the interest of time I'll fix this because, as you say, you're new with GitHub and otherwise you will have to merge my outline structure changes back to your working branch & that can be a tricky business if you haven't done it before...

Kirael12:
Thank you very much @rphair ! I'm still learning Github, it is not part of my skillset, but i love learning new things :) I understand how it works now, and i will be able to produce more Guides and documentation from now (and i will make sure to not capitalise header titles !)

PR #927, opened 2023-01-16
Add a link to community developed smart contract languages

nielstron:
## Checklist

The Cardano smart contract page is the first search result when looking for "Smart contract development on Cardano". It is important to fight the misconception that Cardano allows only Haskell contracts and to show the strength and diversity of the Cardano community. it would hence be great to either list community led alternatives directly or link to the list provided by the community when introducing potential new developers to Cardano.

### DIFF
--- a/docs/smart-contracts/overview.md
+++ b/docs/smart-contracts/overview.md
@@ -24,5 +24,14 @@ In another world, you'd like to build programs, possibly even replace a major co
 - The smart contract is a piece of code that allows a decentralized system to operate.
 
 ## Programming languages
+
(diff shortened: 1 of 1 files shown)


OlofBlomqvist:
I think we should not separate out Aiken or Eopsin in a way that make them seem like second rate languages but include them directly in the list like plutus and marlowe already are on this page :)

nielstron:
Great :) As you can see there are many languages that should be listed but maybe we can reuse the introduction from the Aiken doc to create a short article on each and then have each project add details

PR #930, opened 2023-01-17
Stake Pool Course introduction: grammar, usage, formatting

dschrempf:
Tiny PR, corrects gender bias.

Another thing I noticed (indepdent): Line 153

        2. Slot leader performs the following duties
        3. Collects the transactions to be included in his block.
        4. Includes in his block the random value \(v\) and proof \(π\) obtained from the VRF output.
        5. ...

I think points 3 and the following ones should be further indented. Is that so?


rphair (approved):
Seems reasonable enough.  Have also demoted the subjugate points in the outline above, and make the same corrections to those subordinate points.

FYI this material is likely to go away soon & replaced with the material here (https://github.com/cardano-foundation/developer-portal/tree/revamp-stake-pool-course) but it's good to have this copy corrected in case it's repurposed for use in the new location.

rphair:
Just classifying this as `Typo` instead of `Documentation` because these changes don't really affect the way anything is documented... although the subordination of what the block producers are doing should help people's understanding in some cases.

fill-the-fill (approved):

katomm (approved):

PR #931, opened 2023-01-17
Fix typo in stake pool course

dschrempf:
Tiny PR, fix typo.


rphair (commented):
I think the currently proposed change is correct, but that the sentence it corrects is flawed.  It's not really understandable even if correctly written and as perhaps @rdlrt was noting via emoji it's not even mathematically true.

Maybe in light of https://github.com/cardano-foundation/developer-portal/pull/930#pullrequestreview-1251518724 we should just ignore trivial changes to the whole `docs/stake-pool-course` hierarchy and close the PRs unless there's a significant error in the outdated docs that are likely to be revived as errors elsewhere.

dschrempf:
I see, thank you!

Could you elaborate why this sentence is flawed? It makes (at least gramatical) sense to me.

rphair:
@dschrempf the proposition that a 1% fee pool can provide less delegator rewards than a 2% fee pool is a mathematically useless proposition: being sometimes true and sometimes not true.

dschrempf:
In this case I may miss something. The term "fee" is not mentioned in this section, and I do not see how it is related. The percentage amounts refer to the amount of stake, don't they?

rphair:
I think you are right... but even if the percentages are saturation level rather than the fixed fee, the proposition is still uncertain.  And whether it's ambiguous or simply not very well-written doesn't matter in this case because it still makes the hastily written sentence something I'd be happy (among other editors perhaps) to cast out along with the rest of the obsolescent stake pool course.

So as I said in https://github.com/cardano-foundation/developer-portal/pull/930#issuecomment-1385400653 I really appreciate your attention but also feel this editing rigour would be better applied to the new material in [revamp-stake-pool-course](https://github.com/cardano-foundation/developer-portal/tree/revamp-stake-pool-course) in which you are warmly invited to review the pending material. 😎

rphair (approved):
... TL;DR I'm approving this as a "why not" kind of change (does no harm even if [and especially if] the material we're correcting _is_ obsolescent, with much greater flaws than little grammar mistakes).

katomm (approved):

fill-the-fill (approved):

PR #934, opened 2023-01-20
Add showcase NFTada.io

fill-the-fill:
## Checklist

## Showcase addition

* Title: *NFTada.io*
* Description: *Simple web interface and sophisticated API for automated printing and distribution.*
* Website:   [NFTada.io](https://nftada.io/)
* Source: null
* Tags:
  * `minting`
  * `nftsupport`


rphair (commented):
just a question first @fill-the-fill @cardanians ... is the word "printing" here used as a synonym for "minting"?  That makes sense to me, but could you confirm it if so?  And if they mean something different, can you explain the difference?  This isn't a challenge... I just want to make sure I understand how the headline term is being used. 😎

fill-the-fill:
> just a question first @fill-the-fill @cardanians ... is the word "printing" here used as a synonym for "minting"? That makes sense to me, but could you confirm it if so? And if they mean something different, can you explain the difference? This isn't a challenge... I just want to make sure I understand how the headline term is being used. 😎

I believe it is indeed used as a synonym for minting

rphair (approved):

os11k (approved):

rdlrt (approved):
I was requested to review - reason I did not add my approval:
I growingly do not support minting services that generate tokens for users but do not share the policy keys (same for existing services already merged). This ends up being a dependency on provider, for every action asset owner needs to perform in future.

Approving now to keep the ground equal , as other minting services aren't different and already merged

PR #950, opened 2023-02-07
Remove 'Edit this page' button from autogenerated content pages

fill-the-fill:
## Checklist

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Updating documentation or Bugfix

Remove 'Edit this page' button from autogenerated content pages such as CIPs, Token Registry and Rust Library since it is not a modifiable

PR #947, opened 2023-02-06
Add changelog for January 2023

fill-the-fill:
## Checklist

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Updating documentation or Bugfix

Add changelog for January 2023

PR #946, opened 2023-02-06
Add BALANCE to Showcase

Balance-Analytics:
## Checklist

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Showcase addition

* Title: *Balance*
* Description: *A Blockchain Intelligence app dedicated to supporting the Cardano Blockchain Ecosystem.*
* Website:  [Balance](https://balanceanalytics.io/)
* Source: null
* Tags:
  * `analytics`
  * `pooltool`

PR #942, opened 2023-02-04
Add Finitum Bridge to showcase

inimrod:
## Checklist

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Showcase addition

* Title: *Finitum Bridge*
* Description: *BSC-Cardano token bridge*
* Website:   [Finitum Bridge](https://finitum.io/bridge)
* Source: 
* Tags:
  * `bridge`

PR #952, opened 2023-02-08
Merge PR #919, #910, #922, #929, and more into main

katomm:
Merge PR #919, #910, #922, #929, #928, #926, #931, #930, #925, #933, #932, #935, #934, #936, #937, #946, #947, #942, #950, #943 into main


rphair (approved):
This is basically everything we've done since the last such merge, right?  I guess we've already checked all the code then. 😅

One thing @katomm I don't understand: how come there weren't other batches like this before in the last few months?  What's different about the overall workflow now? 🤔

katomm:
@rphair yes usually I'm pushing staging directly to the main with deployment keys, but there seems to be a problem with my access token or my local setup, and we are a bit outdated on the main branch to wait longer. This is the classy way:  just create a pull request from staging into the main. (good to know in case you want to bring something to the main faster than the normal schedule)

rdlrt (approved):

rphair:
ready to go @katomm :rocket:

PR #978, opened 2023-03-04
Installing cardano-node: update links to binaries

WesleyBatista:
👋👋 Hello there! Welcome. Please follow the instruction below. 

Click the `Preview` tab and select a PR template:

- [Add Showcase](?expand=1&template=showcase.md)
- [Add Builder Tool](?expand=1&template=builder-tool.md)
- [Update documentation or Fix a Bug](?expand=1&template=standard-change.md)

PR #1025, opened 2023-04-13
Update installing-cardano-node.md to ver 1.35.7

gufmar:
## Checklist

update the guide from 1.35.5 to 1.35.7 (download links and min reqs)

### DIFF
--- a/docs/get-started/installing-cardano-node.md
+++ b/docs/get-started/installing-cardano-node.md
@@ -5,6 +5,11 @@ sidebar_label: Installing cardano-node
 description: This guide shows how to build and install the cardano-node and cardano-cli from the source-code for all major Operating Systems
 image: ../img/og/og-getstarted-installing-cardano-node.png
 ---
+:::important
+
+This documents current version is relevant for cardano-node release version [1.35.7](https://github.com/input-output-hk/cardano-node/releases/tag/1.35.7) (Apr-2023)
+
+:::
 ### Overview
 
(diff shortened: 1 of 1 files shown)


rphair (approved):
damn, hard to believe we still don't have the `/latest` "permalink` to the current node version (last time I checked was https://github.com/cardano-foundation/developer-portal/pull/978#issuecomment-1455082948).  @rdlrt I keep asking you since you were also waiting for this result... any progress on that front?

rphair:
wow, great news!  Therefore I would vote for using the `/latest` link instead of the 3 direct links, since the min spec & tooling will go out of date less often than the node version.  The `cabal` and GHC versions also aren't mentioned in the `cardano-node` release pages, so we'd have to go back to original sources to look these things up anyway and/or try compiling it ourselves.

In [the new SPO branch > Cardano Node Installation](https://github.com/cardano-foundation/developer-portal/blob/revamp-stake-pool-course/docs/stake-pool-guide/node-installation-process.md) we linked to the IOG document "in case of difficulties": https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/install.md ... but I can see they haven't even updated the node RAM requirement there to 24GB as they have on the GitHub release page for `1.35.7`, so this may not be considered an original source either.

So if we were double-checking the `cabal` and GHC versions I'm not sure where we'd get them from... and if we're getting them via announcement or a testing process, I believe it would make more sense to insert them here in response to those announcements... while leaving the `/latest` reference the way it is.

gufmar:
happy to commit in another change to make it more general if others agree. Imo the guide could just point to the IOG release page, and recommend to also check the HW and build min requirements there.

katomm (approved):

PR #1050, opened 2023-05-04
Add message about ldconfig

NickJorens:
## Checklist

I wanted to run a preview node using systemd. I followed the exact instructions, but the systemd error output showed an error. After research I found this can be remedied by simply running ```ldconfig``` after setup.


docs/get-started/installing-cardano-node.md
```diff
@@ -515,6 +515,16 @@ cp -p "$(./scripts/bin-path.sh cardano-node)" $HOME/.local/bin/
 cp -p "$(./scripts/bin-path.sh cardano-cli)" $HOME/.local/bin/
 ```
 
+In order to run the cardano-node as a service using systemd you need to run the following command:
```
rdlrt, line 518: Since there are heaps of Linux flavours around:

Simply running ldconfig may not be enough (neither absence of it will guarantee the error, if  LD_LIBRARY_PATH already points to local lib folder).

If using ldconfig, it would be an essential step to ensure /usr/local/lib - or the path where you install the lib to - is part of /etc/ld.so.config (or mentioned in a file in /etc/ld.so.conf.d).

rphair (commented):
I was also pondering this when https://github.com/cardano-foundation/developer-portal/pull/1050#discussion_r1184829276 came in: the statement is literally true for appropriate flavours of Linux, but without background information also makes it sound like this is the only circumstance where you might use `ldconfig`.

You might need it the first time running `cardano-node` from the command line or after adding dependent libraries.  I think the addition would require a more complete explanation of why this is necessary to run `cardano-node` in general and not just for `systemd` while also addressing the issue of what to expect on different platforms.

katomm:
@NickJorens are you finalizing this?

NickJorens:
> @NickJorens are you finalizing this?

I'm sorry, but I wasn't planning on following up on this.

rphair:
OK then @katomm we have to drop this, because the advice offered is only intermittently applicable & not without some explanatory context.

PR #1054, opened 2023-05-09
Add Maestro ISPO Marketplace to Showcase

Vardominator:

## Showcase addition

* Title: *Maestro ISPO Marketplace*
* Description: *Discover top ISPO projects on Cardano.*
* Website:  https://www.gomaestro.org/marketplace
* Source: null
* Tags:
  * `pooltool`


rphair (commented):
I am of two minds about including this, based on 2 separate questions, so I'd like to hear other reviewers' feelings on these in deciding whether this fits into the ethos of the Portal:

1) This tool would be a less conventional member of the `pooltool` category, but this would be OK for me because of how the category is generally used: to make delegation choices.

2) Yet I don't see how this tool could properly represent the ISPO market if all the listed ISPOs are only appearing because they've paid Maestro a $95 or $195 fee.  Since a Dev Portal listing would effectively standardise this is a primary reference for Cardano delegators to choose an ISPO, it would be like a tax on ISPO's going to a single private entity. 🤑

PR #1067, opened 2023-05-24
On-Chain Polls: correct & update explorer links

rphair:
For consistency's sake, set the markup to be exactly the same as currently appearing here (though this changes appearance a bit): https://github.com/cardano-foundation/CIP-0094-polls/blob/main/README.md#existing-polls


rphair:
Deleting this since somehow it has these extra commits in it (will reset my fork's `master` & push this again)

PR #1082, opened 2023-06-09
Add  NFT Creator tool to Showcase

tienna:
### Checklist
- [ ]  I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [ ]  I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [ ]  I have run yarn build after adding my changes without getting any errors.

### Showcase addition
Title: NFT Creator tool
Description: A Website that enables non-tech users to create NFT by themselves with a few simple clicks, it also allows the creator to customize the metadata of NFT, control the policy-id of assets, and use their images for each NFT"
Website: http://nft.easterntownhall.com:8088/ 
Tags:  nftsupport


rphair (commented):
If we had been given (as a common courtesy) write permission to this PR branch, I would correct the broken English in the `description` field and adjust some other things (brevity, and that the link provided is not a "website": it's a page on a subdomain & different port number piggybacking on another web site) in case this is further reviewed while I'm gone (I'll be away most of next week but there are several other reviewers).

Our NFT Support section in the Showcase (https://developers.cardano.org/showcase?tags=nftsupport) has 30 projects and by my check just now 10 of them already are NFT Creators.  I think a "top 10" in any class is a reasonable limit although @tienna listing in the Developer Portal should not be thought of as a contest.

But considering we already have 10 NFT creators with proper web sites... all of them more than just an extra page without documentation, trade dress, or a clear path of responsibility if something goes wrong... then I couldn't accept this submission as it is.  Maybe it has some potential to be worked up into a proper web site if the sponsors can establish something that is uniquely useful about it.

tienna:
Thank @rphair for your valued comment and suggestion. 
I apologize for the incorrect wording in the description as well as for the website redirection as you said
I'm glad if there are other tools with the same functionality as ours listed on the website. I just highlight that our tool allows minting NFT from address pair from the user's wallet, the user can manage asset policyID, and low minting cost (1.18 ADA/NFT).

By the way, may I ask if we include this tool in the builder tool group, what conditions does the tool need to satisfy? we want to share code/docker image for the community to reuse

Regards

PR #1085, opened 2023-06-12
Registering a Pool: add tip for GitHub page metadata

36thchambersoftware:
Adding tip for github users to specifically link to the raw version of the file.

It seems that this issue comes up many times in the stake pool discord help channels. Hopefully specifically stating it in the docs will help some people avoid this problem.


rphair (commented):
I would rather not add this tip because:
1. GitHub as a host for the metadata page is only mentioned as an example & won't apply to everybody.
2. SPOs not having working knowledge of GitHub are likely to be confused by this comment... perhaps wondering if it's applicable to them even if their metadata page _isn't_ hosted on GitHub.

But I'm not against adding it if other reviewers think this suggestion is helpful & I'll go with the crowd on this one.

katomm (changes requested):
The tip is fine for me, but please use "GitHub" instead of "github" to comply with Editorial Style Guide listed on https://developers.cardano.org/docs/portal-style-guide#editorial-style-guide

(hate to be THAT guy)

rphair:
@36thchambersoftware @katomm we've removed GitHub as a suggestion altogether for other reasons in https://github.com/cardano-foundation/developer-portal/pull/1100, so this tip isn't applicable to anything in the source documentation anymore. 🤓

PR #1088, opened 2023-06-14
Add Hydra to builder tools

v0d1ch:
## Checklist

- [x] I have read the [Builder Tool Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/builder-tools.js)

## Builder Tool addition

* Title: *Hydra*
* Description: *Hydra is the layer-two scalability solution for Cardano, which aims to increase the speed of transactions (low latency, high throughput) and minimize transaction cost.* 
* Website:   https://hydra.family/head-protocol/
* Getstarted: https://hydra.family/head-protocol/docs/getting-started
* Tags:
  * `cli`
  * `http`
  * `haskell`


fill-the-fill (approved):

os11k (approved):

ch1bo:
@rphair Can we get some more eyes on this? :)

src/data/builder-tools.js
```diff
@@ -1034,6 +1034,14 @@ export const Showcases = [
     getstarted: "https://docs.gomaestro.org/",
     tags: ["http", "hosted"]
   },
+  {
+    title: "Hydra",
+    description: "Hydra is the layer-two scalability solution for Cardano, which aims to increase the speed of transactions (low latency, high throughput) and minimize transaction cost.",
+    preview: require("./builder-tools/hydra.png"),
+    website: "https://hydra.family/head-protocol/",
+    getstarted: "https://hydra.family/head-protocol/docs/getting-started",
```
rdlrt, line 1042: Get started is not in line with contribution guide

rdlrt (changes requested):

src/data/builder-tools.js
rphair: @v0d1ch @ch1bo I think this is about setting the `getstarted:` field to a page on the Developer Portal itself rather than an external URL.  Providing a page in the [Get Started](https://developers.cardano.org/docs/get-started/) > Builder Tools section would be appropriate for an industry standard like Hydra.  If you can't or don't want to provide such a page, you could set `getstarted:` to `null` and rely on your `website:` link to orient new users.

@rdlrt @katomm I recall there was a place where we advised contributors of the "getstarted" rule, but I don't see it currently in [builder-tools.js or](https://github.com/cardano-foundation/developer-portal/blob/staging/src/data/builder-tools.js) in the contribution guide... have I lost track of it, or has that guideline been eliminated? https://developers.cardano.org/docs/portal-contribute/#add-tools-to-builder-tools

rdlrt: ah - my bad, looks like this was only [the case for operator tools](https://github.com/cardano-foundation/developer-portal/blob/staging/src/data/builder-tools.js#L1164-L1184) (atleast as per rules), I mis-remembered this being applicable to builder tools itself

rphair: yes, I've misremembered the same thing @rdlrt 😅 ... I'm just dismissing the red-flag review so we can merge this :sunglasses:

rphair (approved):

katomm:
@rphair We had that requirement once, but today external get-started pages are okay. The big advantage and that's why I'm personally in favour of get-started pages within the portal: is that they allow the projects to be found via the search function and can thus attract more traffic to their site. (but everyone has to decide for himself)

PR #1096, opened 2023-06-25
Update register-stake-pool.md, lets use protocol.json everywhere

os11k:
## Checklist

Update register-stake-pool.md, lets use protocol.json here as everywhere else instead of params.json, this will introduces a bit more consistency.

### DIFF
--- a/docs/operate-a-stake-pool/register-stake-pool.md
+++ b/docs/operate-a-stake-pool/register-stake-pool.md
@@ -36,7 +36,7 @@ Before starting, make sure you have access to:
 | `kes.skey` | KES signing key |
 | `vrf.vkey` | VRF verification key |
 | `vrf.skey` | VRF signing key |
-| `protocol.json` | protocol-parameters file |
+| `protocol.json` | protocol parameters file |
 
 ## Create JSON file with your metadata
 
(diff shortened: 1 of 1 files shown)


rphair (approved):
I also wanted to update that at the last review of this document, but was of a mixed mind about it.  I think avoiding potential questions like "how are these files different?" is more important than the license to assign an arbitrary name for this file.

os11k:
Sorry, @rphair , for those multiple commits even after I marked this PR as ready for review. I just found a bit more time today than expected and decided to add those bits too. It is still, in my opinion, work in progress, but I will leave it as it is for now. Maybe later I will get a bit more time, but then it will go in a separate PR. Again, sorry for this mess. Cheers!

rphair:
> multiple commits even after I marked this PR as ready for review

😎 no problem, still looks good ✔️

rdlrt (approved):

os11k:
this is a lost one, I promise.... I removed * from IF in UTXP script. I had * for my environment, where * is necessary, because after redirecting output from Docker to local files, it seems to add a /r at the end of the fullUtxo.out file. This is not necessary here, as most users will not be using Docker.

rphair:
fine with me ✔️ and I also realised it would drive me crazy if not removing the `-` (from the machine readable command option) from the human readable English expression "protocol parameters" 🤓

PR #1113, opened 2023-07-27
Proposal to remove Daedalus as favorite

katomm:
Proposal to remove Daedalus as favorite to avoid that new users are persuaded to believe that a full node is the best solution for them.


rdlrt (approved):

rphair (approved):
Yes, I agree this is important.  Also there was a "security" argument presented in favour of Daedalus 3+ years ago that still persists colloquially, and having the listing here helps sustain the belief that the full node setup is "worth it" somehow.

katomm:
Yes, there is no doubt that full nodes are important. We just want to avoid inexperienced people downloading it because it's favorite and then complaining about having to load 75 GB of data etc.

os11k (approved):

PR #1116, opened 2023-07-29
Hardware Requirements update, drop Marketing

rphair:
1 - I noticed some glaringly out of date specs on the "Minimum hardware requirements" page when I was running through the instructions for some other reason (it wasn't a complete review; this just caught my eye).
- This page was added to the "revamp" stake pool course branch in https://github.com/cardano-foundation/developer-portal/pull/826 but I think the timing was between nodes so the bump to 24GB RAM etc. didn't get included.
- I'm updating the statistics as per the currently latest mainnet [release 8.1.2](https://github.com/input-output-hk/cardano-node/releases/tag/8.1.2).

2 - We agreed to drop the marketing page but I see it is still in `staging` and currently [on the live site](https://developers.cardano.org/docs/operate-a-stake-pool/marketing-stake-pool).  If people need a review I will go into the many wrong things on this page and all the reasons why I hate it.  For now I just want to reference our last consensus to drop the page (circa August 2022)
- https://github.com/cardano-foundation/developer-portal/issues/737#issuecomment-1213233260
- https://github.com/cardano-foundation/developer-portal/issues/737#issuecomment-1214249219
- https://github.com/cardano-foundation/developer-portal/issues/737#issuecomment-1217978885

Following the above agreement to drop it from the outline & the [commit history](https://github.com/cardano-foundation/developer-portal/commits/revamp-stake-pool-course/docs/operate-a-stake-pool/marketing-spo.md), I don't think we ever followed through to drop it from the branch & so understandably it got merged in with `staging` along with all the newer stuff. 😅

One nontrivial difference with the last specs is that support for `cardano-node` on Windows and MacOS is clearly indicated (I don't remember when this appeared... if it were up to me I'd say "Don't try this at home, kids").  With some reluctance I believe that means we should indicate Windows & MacOS as well... but only with as much detail as necessary: experienced SPOs would generally go for Linux but we have to support other points of view including the official one.  (I have no idea why they don't include Ubuntu 22.04 since I understand it's in their own testing configuration.)

---

([link to rewritten page](https://github.com/rphair/developer-portal/blob/spo-update-specs-drop-marketing/docs/operate-a-stake-pool/hardware-requirements.md))


docs/operate-a-stake-pool/hardware-requirements.md
```diff
@@ -9,11 +9,11 @@ The latest technical specifications and supported platforms can be found on the
 
 Currently the following specifications are recommended:
 
-- Servers: 1 for block producer node, at least 1 for relay node
+- Servers: 1 for block producer node, at least 1 as relay node
 - CPU: An Intel or AMD x86 processor with two or more cores, at 1.6GHz or faster (2GHz or faster for a stake pool or relay)
-- Memory: 16 GB of RAM
-- Storage: 100GB of free storage (150GB recommended for future growth)
-- Operating system: 64-bit Linux (i.e. Ubuntu Server 20.04 LTS)
+- Memory: 24GB of RAM
```
rdlrt, line 14: Should we convert it to network-wise table (node doco [adopted](https://github.com/input-output-hk/cardano-node/blob/master/doc/getting-started/install.md#prerequisites), tho is outdated)

os11k (approved):

docs/operate-a-stake-pool/hardware-requirements.md
```diff
@@ -9,11 +9,11 @@ The latest technical specifications and supported platforms can be found on the
...
+- Storage: 150GB of free storage (250GB recommended for future growth)
+- Operating system: 64-bit Linux (Ubuntu 18.04+, Mint 19.3+, Debian 10.3+), Windows 8.1+, MacOS 10.13+
```
rdlrt, line 16: If you check supported roles, BP and Relay are  not "Supported" as per release notes for Windows/Mac - thus, IMO maybe good to keep eliminated in docs for "stake pool operators"?
IMG_20230729_224429

rdlrt (commented):
Agree with overall PR, just added a couple of comments about requirement specifics

docs/operate-a-stake-pool/hardware-requirements.md
```diff
@@ -9,11 +9,11 @@ The latest technical specifications and supported platforms can be found on the
 
 Currently the following specifications are recommended:
 
-- Servers: 1 for block producer node, at least 1 for relay node
+- Servers: 1 for block producer node, at least 1 as relay node
```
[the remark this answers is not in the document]
rphair: I thought today about bumping it to 2 instead, and then also about all the pools that are in violation of it, as well as who benefits from the "recommendations" (not the new SPO on a limited budget).  For now I've changed it to 2 in the interest of pushing this through (@os11k please respond if you have any objection).

My heart tells me this number should be a lowest common denominator, which tends to provide flexibility.  If I had to choose between one relay with 12 cores and 2 relays with the "recommended" 2 cores 😨 I would rather have the single relay.  But yes it would be more in keeping with the other changes to use 2 consistently recommended by IOG elsewhere.

docs/operate-a-stake-pool/hardware-requirements.md
rphair: I prefer tables in general, but there's no advantage to putting a table here because
1. The specs are the only thing on the page here, while the documentation page uses it as an introduction to a big load of text: so it works because the specs stand out as a "read this before you continue" to the setup instructions.
2. Particularly the Air Gap node needs to be highlighted for the Mainnet, while irrelevant for the testnets, so this item wouldn't be suited by the table format at all.
3. We have items here that aren't included on the table: these would widen it, make it less responsive, causing horizontal scrollbars and/or cramping in the text-heavy columns.
4. We can't just copy the Markdown for the table here, as we have in other doc sections, because we can see they allow the material to go out of date.

So since the testnet is only different in 2 scales, would you be happy just adding this?  I've updated it in the PR but please @rdlrt @os11k check the values since I have no experience on the testnets:

---

For testnet pools, some requirements are smaller:
- Memory: 4GB of RAM
- Storage: 20GB of free storage
- Air-gapped environment not required

---

This assumes 2 relays are still required on the testnets (I don't think this is strictly true?) and that Windows and Mac nodes are not permitted on testnets (can that really be true?).  I'm OK leaving this to the reader's discretion, but if we're ever going to split hairs it might as well be here. 😜

docs/operate-a-stake-pool/hardware-requirements.md
rphair: You know, I recalled seeing that many times but couldn't find it today... right under the accordion on the node release page itself. 😖 Yes I've removed the Mac & Windows roles (again).

rphair:
Have made some updates.  @rdlrt suggestion of using a table instead would be nice, but I don't think this would work because it would be a very cramped & asymmetrical table.  Link to current version is here (also added to OP): https://github.com/rphair/developer-portal/blob/spo-update-specs-drop-marketing/docs/operate-a-stake-pool/hardware-requirements.md

docs/operate-a-stake-pool/hardware-requirements.md
rdlrt: IMO - 20GB is safe, as current usage on preview/preprod is ~5.5/3.4GB (+account for.near future and OS).
Happy with above

docs/operate-a-stake-pool/hardware-requirements.md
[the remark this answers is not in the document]
rdlrt: >  If I had to choose between one relay with 12 cores and 2 relays with the "recommended" 2 cores

Well worse could get to 1 relay with 2 cores (the idea with 2 is primarily redundancy), also wouldn't want folks could use dev portal reference as "CF recommends 1 relay" 😄  - but ye, as you said it is primarily to keep recommendations consistent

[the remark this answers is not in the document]
rphair: sure, let's not tell them you can just add the BP keys to your 1 relay and _really_ slash your pool running costs 😆

rdlrt (approved):

katomm (approved):

Issue #1117, opened 2023-07-29
Tag Lace wallet as open-source


DominicZ8:
Lace.io wallet went fully open-source in June 2023, according to their blog post: [https://www.lace.io/blog/unleashing-the-power-of-web3-lace-goes-open-source](https://www.lace.io/blog/unleashing-the-power-of-web3-lace-goes-open-source)

This should be updated in the wallet overview:
https://developers.cardano.org/showcase/?tags=wallet

rphair:
Thanks @DominicZ8 for noticing & reporting 🤓 ... fixed by https://github.com/cardano-foundation/developer-portal/pull/1118

PR #1118, opened 2023-07-29
Lace wallet: tag as Open Source & link to repo

rphair:
Fixes https://github.com/cardano-foundation/developer-portal/issues/1117 (cc @olgahryniuk)

### DIFF
--- a/src/data/showcases.js
+++ b/src/data/showcases.js
@@ -1123,7 +1123,7 @@ export const Showcases = [
       "A standardized NFT that developers and users can use to associate an address with a custom and human-readable address.",
     preview: require("./showcase/adahandle.png"),
     website: "https://adahandle.com",
-    source: "https://github.com/input-output-hk/lace",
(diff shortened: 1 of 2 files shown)


rdlrt (approved):
LGTM, I added a commit to remove yarn.lock - but not sure if it was intentional, so reverted back for now. @rphair - please check if applicable

rphair:
thanks @rdlrt - no, I didn't notice the `yarn.lock` was there, and with this file gitignored the whole time I don't see why it would have been in my original commit.  The only thing I can think of is that I recently built at least 1 PR locally... some old one(s) or maybe a new one has `yarn.lock` in its own branch.  Still the best insight I can give people unfortunately is that "it happens every now and then" with a quick procedure for removing it from their PR. 🤔 (it's gone now)

os11k (approved):

katomm (approved):

PR #1120, opened 2023-08-01
Changes on token tools

katomm:
Changes on token tools:
- Add favorite tag for taptools.io, because the platform grew amazingly and they are constantly improving it.
- Remove favorite tag from cardanoassets.com, because it can no longer justified compared to for example taptools
- Add native token tag to cardanoassets.com
- Remove tokentool.io because it is down for some time now


rphair (approved):
I agree TapTools is looking good 😇

rdlrt (approved):

os11k (approved):

PR #1128, opened 2023-08-09
Add --depth 1 for git clone

katomm:
Update portal-contribute.md to use git clone --depth 1 for quicker checkouts


rdlrt (approved):

rphair (approved):
Makes sense.  After over a year of local testing I have still never gone deeper than "`1`" in my local git instance (the bulk of that testing would be pulling PR branches from `upstream`).

PR #1135, opened 2023-08-22
Fix CIP information generation by removing "type" property.

fill-the-fill:
## Checklist

Since according to CIP-0001, `type` property in header was quote:"confusing and not-too-useful" I don't believe we should use it in CIP information at the end of each CIP generated on Developer Portal. In the example bellow CIP-0002 doesn't have `type` property so its displaying `null`.

Currently on Master it looks like this:
Screenshot 2023-08-22 at 15 50 53

I suggest we make it look like this:
Screenshot 2023-08-22 at 15 52 16


rphair (commented):
Another option would be to use:
- `Category` if it's defined in the header, or
- `Type` if not (i.e. the legacy CIP definition).

All CIPs originally had `Type` but then, as a result of following the updated `CIP-0001`, specified a `Category` instead.  All CIPs had either one or the other as a requirement for being merged. Old CIPs that we've updated or will update soon (https://github.com/cardano-foundation/CIPs/issues/389) will have `Type` removed and a `Category` assigned.

The old language that worked for `Type` would actually work well for `Category` preceding the name of the CIP: since the Category is also a modifier (Ledger CIP, Plutus CIP, etc.): https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#categories

After https://github.com/cardano-foundation/CIPs/issues/389 is complete, all CIPs will have only a `Category` field and none of them will have a `Type` field.

rphair:
Just having a look at this again & think these other refinements would be easy to make if fully updating this code for the new CIP-0001 specification:

- the link target for the `Active` status as screenshotted above (right now, [this nonexistent anchor](https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#cip-workflow) on the cip1 page) could be set here: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#status-active
- i.e. either linking each status to its definintion, or all of them to the Statuses section: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#statuses
- likewise, when displaying `Category` instead of `Type`, it could be linked to the Categories table: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#categories

fill-the-fill:
> Just having a look at this again & think these other refinements would be easy to make if fully updating this code for the new CIP-0001 specification:
> 
> * the link target for the `Active` status as screenshotted above (right now, [this nonexistent anchor](https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#cip-workflow) on the `cip-0001` page) could be set here: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#status-active
>   
>   * i.e. either linking each status to its definition, or all of them to the Statuses section: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#statuses
> * likewise, when displaying `Category` instead of `Type`, it could be linked to the Categories table: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0001/#categories

sounds like a good idea, let me update you here when I push a commit with the following changes.

katomm:
Any updates on this? We seem not to move here.

PR #1139, opened 2023-08-31
Minting Native Assets: qualify use of # character

johnnygreeney:
Clarification on use of '#' in this case

## Checklist

 <-- Please fill the boxes with [x] before submitting a pull request --> 

- [ x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [ x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Updating documentation 

As this is a 'getting started' guide, it may be best not to assume every user is experienced with cli. As # is more commonly used as a comment, it makes sense to include short explainer of the syntax expected. For clarity, the norm should be to explain any grey areas on first mention, not repeat everywhere it is used.

### DIFF
--- a/docs/native-tokens/minting.md
+++ b/docs/native-tokens/minting.md
@@ -306,6 +306,8 @@ cardano-cli transaction build-raw \
  --minting-script-file policy/policy.script \
  --out-file matx.raw
 ```
+:::note
+In this case, '#' is not a comment or a 'trim string' as you might expect in a bash script. It is the specific syntax structure expected for the 'tx-in' parameter, ie. --tx-in $txhash<b>#</b>$txix 
 
 :::note 
 In later versions of cardano-cli (at least from >1.31.0) <b>the tokennames must be base16 encoded or you will receive an error</b>


rphair (commented):
I don't think it's necessary to add this explanation because of the long precedent for using `#` characters in UNIX command arguments: not common, but steady across a variety of purposes.

It would also introduce an unsustainable precedent because, if justified to add here, it would also be justified in attaching this comment to every Cardano `Tx` syntax which I believe would be impossible.

johnnygreeney:
It's commonly used as a comment and less so in arguments. https://www.howtogeek.com/439199/15-special-characters-you-need-to-know-for-bash/#comment-or-trim-strings

This is a short 'how to' guide, not documentation.  If using in such fashion only a few times in the guide, it would make sense to explain on first mention only. 

It's justified to explain on first mention, no request or implication it should be explained everywhere.

rdlrt:
> It's commonly used as a comment and less so in arguments. https://www.howtogeek.com/439199/15-special-characters-you-need-to-know-for-bash/#comment-or-trim-strings

"#" following another character is never treated as a comment in POSIX (for instance `echo a#b` => `a#b` , but `echo a #b` => `a` ). Clarifying basic POSIX grammar shouldn't be part of this page, or if done so - should be done consistently (which is difficult since what might be obvious to some might not be to others, the baselines to clarify can get difficult and vague to work with).

PR #1141, opened 2023-09-04
Add Cardano Account Pandas Dumper to Showcase

pixelsoup42:
## Checklist

- [X] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)

## Showcase addition

* Title: cardano_account_pandas_dumper
* Description: cardano_account_pandas_dumper bridges Cardano transaction history with the Pandas data analysis framework, letting users create a report spreadsheet for their wallet's transaction history, including all native assets and many DeFI contracts.
* Website:   https://github.com/pixelsoup42/cardano_account_pandas_dumper
* Source: [<link_to_source_code>](https://github.com/pixelsoup42/cardano_account_pandas_dumper)
* Tags: `analytics`  `metadata`  `nftsupport`  `opensource`


rphair (commented):
@pixelsoup42 looks like a fair amount of work has gone into this & if it works I would be in favour of adding it.

Since this is a new GitHub account, and having no visible track record for this project, we have to scrutinise it more than other submissions.  I read over the code & it looks straightforward except that I feel like there needs to be some explanation (in the code or `README`, not just here) of where the (impressive) list of exchange & DEX addresses came from.

Also: I'm trying to test as a Dev Portal user would, which has been difficult because of the "externally managed" Python environment beginning with Ubuntu 22.04 and other Linuxes following the same standard.  Could you add something to the `README` explaining of how to install & run it through `pipx` or some other way of beating the external management?
```
$ pip install git+https://github.com/pixelsoup42/cardano_account_pandas_dumper
error: externally-managed-environment
```

pixelsoup42:
> @pixelsoup42 looks like a fair amount of work has gone into this & if it works I would be in favour of adding it.
> 
> Since this is a new GitHub account, with no visible track record for this project, we have to scrutinise it more than other submissions. I read over the code & it looks straightforward but I feel like there needs to be some explanation (in the code or `README`, not just here) of where the (impressive) list of exchange & DEX addresses came from.

@rphair thanks a lot for the feedback.

Yeah I created this github account specifically for this tool without being aware of the requirement here for a not-new github account, changing everything would be kind of a hassle and defeat my original intent of keeping this cleanly separated. As you can see, this code does not need any secrets and is about 500 lines of Python, really easy to audit.

I gleaned the list of exchange and DEX addresses from external sources like cardanoscan.io and cexplorer.io, added that to the README.

As I also wrote in the README I'd be really glad to know about an API that would let me fetch this kind of info dynamically, could not find any.

> 
> Also: I'm trying to test as a Dev Portal user would, which has been difficult because of the "externally managed" Python environment beginning with Ubuntu 22.04 and other Linuxes following the same standard. Could you add something to the `README` explaining of how to install & run it through `pipx` or some other way of beating the external management?
> 
> ```
> $ pip install git+https://github.com/pixelsoup42/cardano_account_pandas_dumper
> error: externally-managed-environment
> ```

Yeah using `pipx` is  much better and simpler, I updated the `pyproject.toml` and `README.md` for that.

rphair:
```
pipx install git+https://github.com/pixelsoup42/cardano_account_pandas_dumper
```
- `pipx` won't even find  dependencies unless you specify an additional command option.  People without Python or system admin experience will get that warning and think it's an error.
```
cardano_account_pandas_dumper  --csv_output report.csv <staking_address1> <staking_address2> ...
```
Installing with the first command doesn't make the program accessible in the second command: installed as above it doesn't create an executable anywhere.

I'm ready to approve this if you could first please think of Dev Portal users who will want to use your tool on a vanilla Ubuntu 22.04 installation or mainstream equivalent and have never used `pipx` and maybe not even `pip` or Python before.  Please post again when you have feel you have something that works more-or-less as cut-and-paste.

pixelsoup42:
> ```
> pipx install git+https://github.com/pixelsoup42/cardano_account_pandas_dumper
> ```
> 
> `pipx` won't even find dependencies unless you specify an additional command option. People without Python or system admin experience will get that error and might not see they need to run again with the suggested option to install necessary dependencies.

Sorry, I am  confused, you mention an error in your reply but I don't see the error message in it.

> 
> ```
> cardano_account_pandas_dumper  --csv_output report.csv <staking_address1> <staking_address2> ...
> ```
> 
> Installed as above it doesn't create an executable anywhere.
> 
> I'm ready to approve this if you could first please think of Dev Portal users who will want to use your tool on a vanilla Ubuntu 22.04 installation or mainstream equivalent and have never used `pipx` and maybe not even `pip` or Python before. Please post again when you have feel you have something that works more-or-less as copy-and-paste.

Really sorry, it works just fine for me but I guess that's because I did the one-time `pipx`  `PATH`setup already .

Here is a script that works for me from a clean docker Debian image:
```sh
#!/bin/sh

docker run -i debian <<EOF
apt update >/dev/null 
apt install -y pipx git  >/dev/null 
yes | adduser --quiet test
su test
pipx install git+https://github.com/pixelsoup42/cardano_account_pandas_dumper
pipx ensurepath
bash -l -c 'cardano_account_pandas_dumper -h'
EOF
```

Does the above script work for you ? If so I will add the `apt install -y pipx git`,  `pipx ensurepath`  and re-login steps to the `README`.

rphair:
The first time I `pipx install`'d this tool it only put the script deep in a subdirectory of `~/.local/pipx/venvs` and it only put the dependencies in `~/.local/bin`. I had the apparent misinformation that `pipx ensurepath` just added `~/.local/bin` to `$PATH`, so I didn't run it. 😕

I can see now `pipx ensurepath` does more than that, i.e. exposing the installed packages as executables.  Therefore I think we are all set if you would please add that step to your instructions in the `README`.

pixelsoup42:
OK I have added this to the `README`:


Try to run `cardano_account_pandas_dumper -h`. If it fails with `cardano_account_pandas_dumper: command not found`, you need to run `pipx ensurepath` and open a new terminal window to use the tool.

rphair (approved):
thanks @pixelsoup42 - works well for me and the revised installation instructions should work for users of various OS's and experience levels.

Further reviews (2 more approvals required) might dispute the use of the tags `metadata` and `nftsupport` which are not really the focus of the tool, though they're applied to wallets that are not focused on these 2 things either.

rdlrt:
Sorry - I'll pass for this one, a bit too simple JSON to spreadsheet conversion for my eyes to be a showcase on dev portal (others can approve and merge 👍🏻 )

pixelsoup42:
@rdlrt I don't think the target is devs specifically. There is nothing else that I know of to generate an account statement including all native assets and many DeFi positions, what else would you recommend ? Don't you think that can be useful for just any Cardano user, for tax purposes or other ?

pixelsoup42:
About the `metadata` and `nftsupport` tags: this tool parses transaction metadata, and will display NFT positions, that's why I added these tags.

pixelsoup42:
FWIW, in the showcase I see https://github.com/gitmachtl/cardano-signer which is about 1k lines of JS, and https://github.com/cardano-apexpool/token-registry-api which is about 700 lines of Python and some data.

rphair:
Just also wanted to add that I'm not a routine user of Blockfrost... nor do I really parse JSON routinely for personal reasons... and recently I've found myself using this utility in my own workflow even though there may be better & more portable options.  If that is true for developers I can imagine how it would be of general interest to users.

One improvement I might throw in @pixelsoup42: when you accidentally paste a Cardano address instead of a stake key the warning returned isn't useful to that particular mistake.  Users should be responsible for correcting that, but perhaps some logic could be added to detect common errors like that & help the user a bit... the current response is to validate your API key from Blockfrost which could send new users down the wrong path. 🤔

pixelsoup42:
@rphair thanks a lot for the feedback. I have pushed a new version that adds staking rewards to the output, and supports graphical output with a timeline for each asset.
I have also improved error handling as you suggested, messages should be much more helpful now.
You can get the updated version with:

```sh
pipx upgrade cardano_account_pandas_dumper
```

rphair:
thanks for your attention @pixelsoup42 ... I'm still in favour of adding it 😎✅

rphair:
Just had to upgrade this tool to keep using it (new Blockfrost endpoint?); maybe like the also Python hosted `youtube-dl` it could include a suggestion that an upgrade might fix the problem?

```
$ cardano_account_pandas_dumper --csv_output account.csv stake1...
```
`Failed to read data from blockfrost.io: HTTPSConnectionPool(host='cardano-mainnet.blockfrost.io', port=443): Max retries exceeded with url ... (Caused by NameResolutionError(" ... Failed to resolve 'cardano-mainnet.blockfrost.io' ...`
```
$ pipx upgrade cardano_account_pandas_dumper
upgraded package cardano-account-pandas-dumper from 2023.2.0 to 2023.2.3
```

BTW I'm still in favour of adding this to the Dev Portal, based on my own personal attachment to it.  Maybe it's because I'm too lazy to formulate a Blockfrost query or massage the results but I do find myself using this regularly.  cc @katomm @os11k @fill-the-fill

rphair:
p.s. @rdlrt I can see this tool is continuing to improve based on

pixelsoup42:
@rphair Thanks for your feedback and support. I have added a "Troubleshooting" section to the README with your suggestions.

PR #1162, opened 2023-10-09
Cardano Relay Configuration: fix & clarify instructions

### LINKED ISSUE #1140: Cardano Relay Configuration: startTestNode.sh not portable, instructions incomplete
While following the instructions for [configuring a testnet node](https://developers.cardano.org/docs/operate-a-stake-pool/cardano-relay-configuration) I found the [startTestNode.sh](https://developers.cardano.org/docs/operate-a-stake-pool/cardano-relay-configuration#creating-startup-scripts-and-services) script had the following problems.

- is not clear that the user should replace `$HOME` by its actual value on the script. This is clear for the `cardano-testnode.service ` file, but I believe isn't for the script since is mentioned in the context of the latter.
- by following the instructions you get the file `$HOME/.local/bin/cardano-node` but not `/usr/local/bin/cardano-node`, which is what the script relies on. There should be a step where the user is told to set up `/usr/local/bin/cardano-node` or instead that line could be changed by `$HOME/.local/bin/cardano-node` 
- While running the script I got the error `error while loading shared libraries: libsecp256k1.so.0: cannot open shared object file:` which was caused by the `LD_LIBRARY_PATH` and `PKG_CONFIG_PATH` variables not being properly set, since running the `systemctl` unit has no access to `$HOME/.bashrc` where it is actually configured by following the instructions. My solution was to set them in the script with the same values it has on `$HOME/.bashrc`
- Also got the following error running the script `Conway related : There was an error reading the genesis file: $HOME/cardano-testnet/conway-genesis.json Error: $HOME/cardano-testnet/conway-genesis.json: openBinaryFile: does not exist (No such file or directory)`. My solution was doing `curl -O -J https://book.world.dev.cardano.org/environments/preprod/conway-genesis.json`

sanskys:

Documentation Update to Fix #1140
1. $HOME is being replaced by <$HOME> to reflect real value
2. ".local" is being replaced by "local"
3. conway-genesis.json file has been added


docs/operate-a-stake-pool/cardano-relay-configuration.md
```diff
@@ -37,6 +37,7 @@ curl -O -J https://book.world.dev.cardano.org/environments/preprod/topology.json
 curl -O -J https://book.world.dev.cardano.org/environments/preprod/byron-genesis.json
 curl -O -J https://book.world.dev.cardano.org/environments/preprod/shelley-genesis.json
 curl -O -J https://book.world.dev.cardano.org/environments/preprod/alonzo-genesis.json
+curl -O -J https://book.world.dev.cardano.org/environments/preprod/conway-genesis.json
```
rphair, line 40: 💯

docs/operate-a-stake-pool/cardano-relay-configuration.md
```diff
@@ -102,6 +103,14 @@ Now create the start script startTestNode.sh
 nano startTestNode.sh
 ```
 
+:::note
+
+Replace <$HOME> with their values of the environment variable. To find the values use following commands:
+
+    echo $HOME
+
+:::
```
rphair, line 112:
```suggestion
:::note
Replace <$HOME> with local values of the `$HOME` environment variable on each mahine. To find those values, run `echo $HOME`.
:::
```

docs/operate-a-stake-pool/cardano-relay-configuration.md
```diff
@@ -111,13 +120,13 @@ PORT=6000
 # 0.0.0.0 listens on all local IP addresses for the computer
 HOSTADDR=0.0.0.0
 # Set a variable to indicate the file path to your topology file
-TOPOLOGY=$HOME/cardano-testnet/topology.json
+TOPOLOGY=<$HOME>/cardano-testnet/topology.json
```
rphair, line 123: The point of the angle brackets is to force the user to substitute the `$HOME` variable, but has the side effect of bombing out the script with an erroneous file redirect if the user doesn't substitute that string as requested.

At least in the prior version of the code there is a change `$HOME` will be defined correctly.

If such a change needs to be made I would recommend an undefined literal like `insertHomeDirectoryHere` which catches the user's attention and would also be present in a meaningful error message if the user doesn't update the directory.

docs/operate-a-stake-pool/node-installation-process.md
```diff
@@ -216,18 +216,18 @@ We can now build the `Haskell-based` `cardano-node` to produce executable binari
 cabal build cardano-node cardano-cli
 ```
 
-Install the newly built node and CLI commands to the $HOME/.local/bin directory:
+Install the newly built node and CLI commands to the $HOME/local/bin directory:
```
rphair, line 219: I sincerely believe this change is very wrong since `~/.local` has become a Linux fixture, especially with KDE and GNOME, while ever since the early days of UNIX `~/local` would in fact be discouraged... it only appears without a dot in often locally writable directories like `/usr/local`.

rphair (commented):
I don't think this page is maintainable for reasons @rdlrt already gave in https://github.com/cardano-foundation/developer-portal/issues/1140#issuecomment-1704571840 ... and therefore I believe most of the content on this page should be replaced with links to the relevant sections on the canonical page [Installing cardano-node and cardano-cli from source](https://developers.cardano.org/docs/get-started/installing-cardano-node/).

As long as the redundancy in the page remains, though, we might as well allow some cleanup.  Some feedback about the proposed changes:

sanskys:
if a cleanup is planned, should i close this PR. Alternatively, i could make the suggested changes as an interim solution?

rphair:
@sanskys I would want to hear what other reviewers think about this, but I firmly believe the only suitable change to make to this page is to eliminate the redundant material, replacing it with a links to the more canonical Dev Portal source.

Perhaps the page & workflow could be rephrased like: "Now that you've created a node according to these (link) standard instructions, here's how to configure it as a relay for your stake pool."

In any case I believe it would be a huge contribution if you could revise your fixes to strip down most of the content & only include those instructions that are unique to SPOs & their relays, without the general node maintenance & operation.  What do you think of that plan? 🤓

sanskys:
> @sanskys I would want to hear what other reviewers think about this, but I still firmly believe the only suitable change to make to this page is to eliminate the redundant material, replacing it with link(s) to the more canonical Dev Portal source.
> 
> Perhaps the page & workflow could be simplified with wording like: "Now that you've created a node according to these (link) standard instructions, here's how to configure it as a relay for your stake pool."
> 
> In any case I believe it would be a huge contribution if you could revise your fixes to strip down most of the content & only include those instructions that are unique to SPOs & their relays, without the general node maintenance & operation. What do you think of that plan? 🤓

@rphair thats fine for me. But since the cleanup would take some time, I would subject to merge these changes and open a new Issue. I would also like to hear more voices before starting. It should not be that after the cleanup everyone finds it a mishmash :)

rphair:
sure thing @sanskys - I'll approve this for merge if it gets approval from 2 other reviewers. 🙂

rdlrt (approved):

os11k (approved):

rphair (approved):

docs/operate-a-stake-pool/cardano-relay-configuration.md
rphair: @sanskys this problem will go away if we proceed as tentatively agreed in https://github.com/cardano-foundation/developer-portal/pull/1162#issuecomment-1787925058.

docs/operate-a-stake-pool/node-installation-process.md
rphair: fixed in https://github.com/cardano-foundation/developer-portal/pull/1162/commits/8b0f3f1d503b67e7701e116f20397a7857ab0b77

rphair:
I've backed out all the superflous & contraindicated changes from the standard `~/.local` to the unusual `~/local`.  The only remaining changes here are the vital Conway genesis file & the instructions about updating the "script" which also adds the error-prone angle brackets.  Due to the latter issue this should be considered an interim edit, pending I hope folding most of this page into links to more canonical instructions 🤓

rphair (approved):

PR #1173, opened 2023-11-22
Add developer blog mainstreet

weqanhet:
## Checklist

Add developer blog mainstreet


katomm (approved):
Looks good, although I'm not a fan of the tags. To be useful we should avoid using unique tags like "mainstreet" that do not appear in other articles.

fill-the-fill (approved):

PR #1177, opened 2023-12-07
Transaction Metadata overview: Correct ranges of 64bit integers

bjornkihlberg:
## Checklist

- [x] ~I have run `yarn build` after adding my changes **without getting any errors**.~
  No, I'm only changing the readme.

Correct ranges of 64bit signed and unsigned integers.

I'm assuming it should be a 64bit integer, in which case the ranges are currently incorrect. Here's a stackoverflow thread with an answer with 87 upvotes strengthening my argument if you don't believe me.


rphair (commented):
The range after `Integers in the range` is obviously impossible with a 64-bit value... but I think we need a definitive reference or schema saying whether this value is signed or unsigned.

That section, even after your change, implies the integer metadata is a signed value... while the earlier text "`between 0 and` ..." for your first change suggests that it is unsigned.  This was also a problem with the original text and therefore also indicates a need to reference the source material in this PR and probably also the document itself.

os11k:
Thank you for your PR! I would like to add, even if you "only changing the readme" it is mandatory to run `yarn build`, that why it is in a checklist. Stackoverflow is nice source of information, but more hard evidences would be necessary, imho.

PR #1213, opened 2024-02-18
fix: correct url for installing cardano node from source code

ShawnTseng:
## Checklist

Updating documentation for installing a node from source code.

`curl -s https://api.github.com/repos/input-output-hk/cardano-node/releases/latest`
gets response below and it does not contain `tag_name` for checkout:

```
{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/188299874/releases/latest",
  "documentation_url": "https://docs.github.com/rest/guides/best-practices-for-using-the-rest-api#follow-redirects"
}
```

So I just adjust the link from .../input-output-hk... to .../IntersectMBO/...


rdlrt (approved):

rphair (approved):
I guess we need to run a search to check for all these `input-output-hk` > `IntersectMBO` updates: both in the API & web URLs.  We got a list of repos in an earlier PR & I'd been meaning to do it and then the time disappeared.  I think it would save time in the long run to update all of these at once. 🤔

ShawnTseng:
BTW, I still can't install node successfully since there are some steps can not work as expect(Debugging).
If replacing all repo's name at one time, We might need to check all guides can work as usual.

rphair:
@ShawnTseng I agree that testing our instructions should be done regularly, but in https://github.com/cardano-foundation/developer-portal/pull/1213#pullrequestreview-1887194099 I was only talking about changing the URLs for the repos that have moved from one account to another.  In almost every case they've been redirected to the new location, so this shouldn't affect whether the URLs are accessible or whether the software will build.

PR #1215, opened 2024-02-19
Fix overview_nfts.svg

katomm:
Fix overview_nfts.svg:
```
[WARNING] The image at "/Users/tommy/Sites/developer-portal/static/img/nfts/overview_nfts.svg" can't be read correctly. Please ensure it's a valid image.
unsupported file type: undefined (file: /Users/tommy/Sites/developer-portal/static/img/nfts/overview_nfts.svg)
```


rdlrt (approved):

rphair (commented):
@katomm I'll approve this for any reason you insist is valid, but:

1. When I build this locally, I don't get the message quoted in the OP.  I can't find this warning in the GitHub CI online builds either.  What tooling is causing this warning?  (FYI I'm using Yarn `3.3.1`)
2. The old file displays properly in all the browsers I checked, as well as Linux image viewers, so I'm sure its SVG is parseable.
3. The new file has smaller fonts, and smaller text areas it looks like... so it's generally less readable and is truncated in several places: enough that several of the headings in the new image are not complete.

Can you double check the image & maybe explain a bit more why it's being replaced? 🙏

katomm:
You are right, I didn't notice the truncated text :-/

To reproduce, you need to open https://developers.cardano.org/docs/native-tokens/ and look at the terminal window.

katomm:
I am converting this to draft for now

PR #1253, opened 2024-04-25
Update installing-cardano-node.md

mplomin:
Final build of Cardano-node-8.9.2 fails with recommended GHC version 8.10.7 due to inability to handle name shadowing in Dif-0.5.

Changed to recommend GHC version 9.4.8, which results in a successful build.


mplomin:
changing to version 9.4.8 only resolves errors in building Dif-0.5, but it creates errors when building Ouroboros.

PR #1280, opened 2024-06-26
Redirect old SPO course links to moved or related pages

### LINKED ISSUE #1258: Broken links silently redirect to unrelated pages
We maintain some docs for our tooling that reference the developer portal.
We have a CI check that ensures our links do not point to 404 pages.
The problem is that some pages were deleted on the portal, but the redirects do not let the user see their equivalents.

E.g. https://developers.cardano.org/docs/stake-pool-course/handbook/keys-addresses/ which used to point to a guide on key generation redirects to https://developers.cardano.org/docs/operate-a-stake-pool/ now.

Ideally the redirects should point to the same articles if they are moved, or respond with a 404 so that we can see that some of our links are outdated.

rphair:
Fixes #1258.  There weren't any "broken" links (as identified in the issue) on the Dev Portal itself; which can still be tested with something like
```
find blog changelog docs examples scripts static -L -type f -print0 | xargs -0 grep stake-pool-course
```
... only an obsolete anchor on the site Overview page and the obsolete section still listed in the file tree on `portal-contribute.md`.  Therefore the affected links are 1) only in the SPO section (unless contradicted in #1258) and 2) are fixed by installing sensible redirects here (since there will never again be actual content at those pages).

There aren't many subjective decisions here, but one of them was redirecting `/docs/stake-pool-course/handbook/setup-firewall` to the beginning of `/docs/operate-a-stake-pool/hardening-server/` rather than the Firewall section within that page (due to the old & new contexts being slightly different).  Any further suggestions please feel free to mark them up in your review.

Tested in local build ✅ along with selected redirects copy/pasted as they would be when appearing in external sites.

cc @klntsky @katomm

### DIFF
--- a/docs/get-started/overview.md
+++ b/docs/get-started/overview.md
@@ -25,7 +25,7 @@ Besides cryptographic research, there was game-theoretic research, identity-mana
 ## What you need to bring
 To get the most out of the Cardano Developer Portal, you should  have programming experience and a basic understanding of blockchain concepts of Cardano such as [UTxO](technical-concepts#unspent-transaction-output-utxo), [transactions](technical-concepts#transactions), [addresses](technical-concepts#addresses), [key derivation](technical-concepts#key-derivation), and [networking](technical-concepts#networking). 
 
-If you are unfamiliar with these terms, start with [technical concepts](technical-concepts), and you can complete the [stake pool course](../operate-a-stake-pool/#stake-pool-course) afterward. It will also help you understand basic concepts, even if you don't want to run a stake pool. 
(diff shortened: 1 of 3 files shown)


rphair:
@sanskys since you worked a lot on this material maybe you would also review this & see if you can come up with any improvements.

sanskys:
> @sanskys since you worked a lot on this material maybe you would also review this & see if you can come up with any improvements.

sure.. will have a look into it in coming weeks.

rdlrt (approved):

os11k (approved):

rphair:
> sure.. will have a look into it in coming weeks.

thanks @sanskys ... then it's probably best for us to merge this now & if you want to make adjustments later, you can do another PR or just send me the adjustments I I'll submit them.

fill-the-fill (approved):

Issue #1327, opened 2024-09-25
404 on cardano-cli link


bezirg:
The link should be updated from

https://github.com/IntersectMBO/cardano-node/tree/master/cardano-cli#cardano-cli   (404 now)

to

https://github.com/IntersectMBO/cardano-cli

![image](https://github.com/user-attachments/assets/ffb12d30-657c-4a95-93db-978f6c000221)

rphair:
@katomm @CarlosLopezDeLara I think this is important to justify its own PR to fix it, rather than waiting to search for similar problems.

@bezirg this will be fixed by https://github.com/cardano-foundation/developer-portal/pull/1329 as soon as it's merged into `staging` and then migrated to the production branch soon after.

PR #1329, opened 2024-09-25
Update cardano-cli link in Builder Tools

rphair:
Fixes https://github.com/cardano-foundation/developer-portal/issues/1327 (thanks for noticing this, @bezirg).

@CarlosLopezDeLara I know the heading for the first text section on this page has changed several times, but I do think using the anchor to jump to that text (past all the GitHub files) is important for new arrivals.  If that first heading changes again, this anchor won't work anymore and it will just land at the beginning of the GitHub page... which it would do anyway if the anchor were left off the link.


CarlosLopezDeLara (approved):

katomm (approved):
good catch

gufmar (approved):
LGTM

PR #1359, opened 2024-11-21
Security Best Practices: fix relative pathnames

rphair:
Fixes problem reported [here on Discord](https://discord.com/channels/875366566944968785/900041546160799784/1309034615377494076).

TL;DR: Some link syntaxes seem browser dependent & this is an attempt to standardise on a known working syntax for some material I wrote before some standards about inter-document linking were more rigorously applied.

For the link to `Installing the node from source` [here](https://developers.cardano.org/docs/get-started/air-gap/#install-cardano-cli), on the live site I've seen it open correctly with the URL into a new window but fail when simply clicking on the link.
* To me this suggests something on the front end is interpreting the links (and likely not correct for me & my JS blocker) and sometimes getting it wrong when the Docusaurus source begins with `..` or a relative pathname that doesn't begin with `docs` (although `/docs` seems to work just as well).

I've standardised the Security Best Practices pages which I wrote long ago & which had a hodgepodge of syntaxes, including the Air Gap link to the "Install from source" page that was reported broken above.  I've concluded linking to `docs/.../<filename>.md` is canonical because output from this comment generally shows that syntax:

```shell
grep '](' `find docs -type f -name \*.md` | grep -v http | grep -v static.img | grep -v '(#' | grep -v cardano-improvement-proposals | grep '(.*)'
```

after repeated, recent edits by @CarlosLopezDeLara @gufmar on these files which have the most intra-Portal links: and so they are especially welcome to check the syntax changes here.  (There are some other edits to bring the content up to date.)


rdlrt (approved):

katomm:
Robert I would approve this without issue, but I'm not getting what we are gaining with using "docs/something" instead of "/docs/something" as the latter will always work, even if files are moved in the paths or show via a different structure (for example tags, or other search/categories paths. And I believe this was the reason why we ended up with "/docs/something" in (mostly) all links.

katomm (approved):

weqanhet (approved):
👍🏽

Issue #1402, opened 2025-02-07
Search Field on Showcase Page Not Clickable in Desktop Browser


lodl:
Description:
On the page https://developers.cardano.org/showcase/, the search field displayed below the search filters is not responsive in a desktop browser. Users are unable to click into the field and type any text.

Interestingly, the field becomes clickable and accepts input only when the browser window is resized to a smaller (mobile) viewport.

Possible Causes:

- An overlapping (overlay) element might be blocking the search field.
- There may be issues with state and DOM synchronization in the React application

Steps to Reproduce:

- Open https://developers.cardano.org/showcase/ in a desktop browser (e.g., Chrome, Firefox).
- Attempt to click the search field located in the top right area (under the search filters); note that it is unresponsive.
- Resize the browser window (reduce its width) so that the mobile layout is activated.
- The search field then becomes clickable and accepts text input.

Expected Behavior:
The search field should be active and responsive in desktop view, allowing users to click and enter text without needing to resize the browser window.

Additional Information:

- Browser: Brave Version 1.74.51 Chromium: 132.0.6834.160
- Operating System: MacOS

katomm:
From a quick test the problem seems to be connected to the ada logo overlay, it will only be blocked where the A covers the input box.

PR #1403, opened 2025-02-07
Fix issue Search Field on Showcase Page Not Clickable in Desktop Browser

lodl:
**Fixes Issue:**  
[[Cardano Foundation Developer Portal Issue #1402](https://github.com/cardano-foundation/developer-portal/issues/1402)](https://github.com/cardano-foundation/developer-portal/issues/1402)  

**Summary of Changes:**  
This PR resolves an issue where the background pseudo-element in the `backgroundAda` class was blocking click events on underlying elements such as the search bar. To fix this, the following code was added to the end of the file `src/components/BackgroundWrapper/styles.module.css`:

```css
.backgroundAda:before {
  pointer-events: none;
}
```

**Details:**  
- This update ensures that the pseudo-element no longer interferes with click events.  
- If the search bar is placed within a container using the `.backgroundAda` class, this change allows user interactions to pass through seamlessly, regardless of the size or positioning of the background graphic.

**Testing:**  
- Verified that the change allows clicks on the search bar without affecting the visual styling of `.backgroundAda`.  
- Confirmed no regressions in other components using the `.backgroundAda` class.

### DIFF
--- a/src/components/BackgroundWrapper/styles.module.css
+++ b/src/components/BackgroundWrapper/styles.module.css
@@ -25,3 +25,7 @@
 [data-theme="dark"] .backgroundAda:before {
   color: #98b7f2; /* dark mode color for the A */
 }
+
+.backgroundAda:before {
+  pointer-events: none;
+}


weqanhet (approved):
LGTM

rphair (approved):
I can't reproduce the problem on Linux and/or Firefox (I use both) but it seems sensible... and I've tested it in local build ✅ and it works as well for me as it did before.

rdlrt (approved):

PR #1370, opened 2024-12-04
Add Nio to showcase

openVietAnh:
👋 Hello there! Welcome. Please follow the steps below to tell us about your contribution.

1. Please complete a Checklist
2. Fill in all sections of the template
3. Click "Create pull request"

## Checklist

 <-- Please fill the boxes with [x] before submitting a pull request --> 

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 
- [x] I have not committed any changes to `yarn.lock` (or have [removed these changes](https://developers.cardano.org/docs/portal-contribute/#i-committed-yarnlock-to-my-pr-branch-what-do-i-do)).

## Showcase addition

<-- Provide information for every bullet in the list below. The tags you select must match the tags in your changes to the showcase.js  -->

* Title: *Nio*
* Description: *Track all assets in wallets and locked in Cardano DeFi smart contracts. Effortlessly monitor your assets and whale wallets*
* Website:   [nioapp.io](https://nioapp.io/)
* Source: 
* Tags:
  * `analytics`
  * `defi`

PR #1438, opened 2025-04-10
Showcase: add Accounting category

rphair:
... according to tentative definition in https://github.com/cardano-foundation/developer-portal/pull/1370#issuecomment-2779256962, broadened from initial suggestion of a category to include "portfolio trackers" so this could also include specialised analytics for financial purposes beyond the holdings of an individual (e.g. a sidechain or token).

Rationale: anything that can be done to track a user's wallet in economic value can also do the same thing for collective resources... but if we have 2 separate categories we risk fragmentation & "too many categories" on the Portal.


weqanhet (approved):
Great addition!

os11k (approved):

katomm (approved):
Great addition.

PR #1457, opened 2025-05-08
Fix search bar on Builder Tools and Showcase

lodl:
This PR addresses an issue where the search bar on the /showcase and /tools pages would lose focus after each character input. This occurred because the URL was updated immediately with every keystroke, causing the component to re-render and, consequently, lose focus.

**Changes Made:**

_Debounced Search Functionality:_

The URL update (and thus the triggering of the filtering logic) is now debounced by 300 milliseconds using lodash/debounce. This means the URL is only updated after the user has paused typing for a short period.
The input in the search field itself is still displayed immediately to have a smooth user experience.

Affected files:

- /src/pages/showcase/index.js
- /src/pages/tools/index.js

_Improved Focus Restoration:_

The restoreUserState function is now called within a setTimeout(..., 0) inside the useEffect hook of the useFilteredProjects function.

This ensures that focus restoration occurs only after all DOM updates and state changes triggered by the URL modification have completed, reliably resolving the focus loss issue.

Affected files:

- /src/pages/showcase/index.js
- /src/pages/tools/index.js

### DIFF
--- a/src/pages/showcase/index.js
+++ b/src/pages/showcase/index.js
@@ -18,6 +18,7 @@ import PortalHero from "../portalhero";
 import { toggleListItem } from "../../utils/jsUtils";
 import { SortedShowcases, Tags, TagList, Showcases } from "../../data/showcases";
 import { useHistory, useLocation } from "@docusaurus/router";
+import _debounce from 'lodash/debounce';
 import styles from "./styles.module.css";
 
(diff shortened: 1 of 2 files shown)


katomm (approved):

gufmar (approved):
LGTM

rdlrt (approved):

PR #1458, opened 2025-05-09
Add stuff.io to showcase

lodl:
This Pull Request is to add stuff.io to the Developer Portal Showcase.
Relevant tags chosen are: marketplace.

- [x]  I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x]  I have run yarn build after adding my changes without getting any errors.
- [x]  I have not committed any changes to yarn.lock (or have [removed these changes]

(https://developers.cardano.org/docs/portal-contribute/#i-committed-yarnlock-to-my-pr-branch-what-do-i-do)).

### DIFF
--- a/src/data/showcases.js
+++ b/src/data/showcases.js
@@ -27,6 +27,14 @@ export const Tags = {
     color: '#e9669e',  // Vibrant Pink
   },
 
+  // Accounting
+  accounting: {
+    label: "Accounting",
+    description: "Tools providing specialised analytics for financial purposes, including portfolio tracking.",
+    icon: null,
+    color: '#85BB65',  // Dollar Bill Green
+  },
+
   // Analytics
   analytics: {
     label: "Analytics",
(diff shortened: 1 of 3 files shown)


weqanhet (approved):

gufmar (approved):
LGTM

katomm:
Is this distinct enough from the book.io entry or should it be replaced with book.io?

goatmustardwildflower:
> Is this distinct enough from the book.io entry or should it be replaced with book.io?

It's distinct from Book.io. As of now, Book.io focuses on selling eBooks & audiobooks on chain, while Stuff.io is selling other digital media, like music and videos. 
Stuff.io is likely to become the umbrella brand under which Book.io will be active later on. If that happens, both entries will need to be revisited.

rphair (approved):
@goatmustardwildflower please let us know and/or submit an update PR if & when the 2 web sites are merged into one: with a tip that one is the "umbrella".

@katomm I'll defer to you about merging this in case it still isn't clear that these 2 web sites are distinct enough.

PR #1460, opened 2025-05-09
Fixed instructions to register a stake with an SRV record.

coot:


rphair (approved):
thanks @coot - I think this is much more clear than the last iteration.

PR #1462, opened 2025-05-12
Getting cardano-node: Remove shell prompt characters

weebl2000:
Remove shell prompt characters so commands can be properly copy-pasted.


rphair (approved):
Only these few out of the many command sequences have these, so I believe it does make sense to remove them.

CarlosLopezDeLara (approved):
LGTM

katomm (approved):

PR #1483, opened 2025-06-09
SPO Generating wallet keys: add stake verification key to payment address

DaveBelv:
👋 Hello there! Welcome. Please follow the steps below to tell us about your contribution.

1. Please complete a Checklist
2. Fill in all sections of the template
3. Click "Create pull request"

## Checklist

 <-- Please fill the boxes with [x] before submitting a pull request --> 

- [ X] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [ X] I have run `yarn build` after adding my changes **without getting any errors**. 
- [ X] I have not committed any changes to `yarn.lock` (or have [removed these changes](https://developers.cardano.org/docs/portal-contribute/#i-committed-yarnlock-to-my-pr-branch-what-do-i-do)).

The command for generating the payment.addr needs the stake.vkey otherwise generated address will not delegate to the stake address.


docs/operate-a-stake-pool/generating-wallet-keys.md
```diff
@@ -62,12 +62,14 @@ The next step would be to generate a wallet address for the payment key `payment
 ```
 cardano-cli address build \
     --payment-verification-key-file payment.vkey \
+    --stake-verification-key-file stake.vkey \
```
rdlrt, line 65: What is the reason for the change? If including stake key, this will no longer be just a "payment address".
This particular address does not need to stake directly, it is used to pay for deposits and transactions - on testnets, it is recipient of th (unless there is a reference elsewhere - if so, please add that as context)

rdlrt: That's fair, I think the flow loses the context when it assumes `payment.addr` is linked to stake account [here](https://github.com/cardano-foundation/developer-portal/blob/bebaa9f57f5b59fb55a1d3be50ccf7caac86be5a/docs/operate-a-stake-pool/register-stake-pool.md?plain=1#L30) without clarifying how it was created.

Overall the language is a bit inconsistent on portal:
- Payment address would typically be associated to payment key hash (which is also previously referred to as `Enterprise` address).
- Many guides call the addresses with linked stake key as base address.
- I believe the intent here was to get initial address populated with faucet funds, and from there on - go and create the actual wallet keys to be used for stake pool operations, which - IMO - would be a better step than directly using single address for everything. If so, we should potentially update `payment.addr` reference on `register-stake-pool.md` and replace it with `base.addr`, we can include a section in this page to create base.addr as well (the source of funds for existing transaction definitions would remain the same).

Ofcourse, that goes beyond the original scope of what this PR was meant to address, so happy to just go with approval for now - if we'd like to tackle this later

rphair (dismissed):
Approving as first order approximation suggested in https://github.com/cardano-foundation/developer-portal/pull/1483#discussion_r2136830218.

docs/operate-a-stake-pool/generating-wallet-keys.md
rdlrt: My vote would be keep wording here referring to payment address and create a base address, while using base address as reference on register-stake-pool.md

DaveBelv: Happy to edit the PR to reflect that. I think it will also help understanding of the stake pool (primarily the active pledge) and the different between the accounts. 

I'll also add a note that other guides (atleast coincashew) might use the base address over the payment here (or when the guide creates the base)  to help with the overall understanding across the user base.

@rdlrt can you please mark as work required just to be clear we are heading down this path. Thanks!

rdlrt (changes requested):
As per discussions, aligned to update wordings to use payment addresses for payment keys (as now) and create a small heading for creating base address.
Additionally, the reference needs to be updated on successive pages for stake pool registrations

aldo555:
@rphair We are doing some testing with vercel but currently there are no production deployments there so don't mind it.

0xBora:
Since the underlying content has shifted quite a bit, the PR has drifted drastically. From what I can see, this got fixed in the operator [docs overhaul](https://github.com/cardano-foundation/developer-portal/pull/1780). The page moved to docs/operators/block-producer/generating-wallet-keys.md, and the address is now built with both the payment and stake keys, with a line explaining that's what lets funds sent there accrue rewards:

https://github.com/cardano-foundation/developer-portal/blob/616a6adef1acbc4e1817beeb120c2dbb97d1c4a0/docs/operators/block-producer/generating-wallet-keys.md#L51-L60

Since the change is in and this branch targets the old path (now conflicting), closing it.

Issue #1485, opened 2025-06-15
🔄 Portal 2025 - 1.1 - The Purge - Builder Tools


0xBora:
## Statement

Builder Tools section has 107 tools listed, but a bunch of them don't work anymore. I've been going through each and everyone of them and found tools with dead websites, archived GitHub repos, and documentation that points to APIs that probably don't exist anymore.

This can easily create a terrible experience for new developers. They click on a tool, the website is broken, they try another one, the repo hasn't been touched in 3 years. By the third broken tool, their ideas about the ecosystem might already be settled falsefully.

This also means most used tools in 2025 are less visible, while outdated tools can be favorited. 

## Proposal

With some builder tools its easy to tell if they are not maintained anymore (last commit 2 or 3+ years ago) or have unanswered GitHub issues. But sometimes its not that easy.

So this needs a systematic approach or basic methods to determine if an item can be called inactive. With this idea in mind, I went down the trenches and have mapped out the 107 builder tools and labeled them, so the idea is not left in abstract.

- **Remove the obviously broken stuff**: If a website returns 404, the GitHub repo is publically archived/inaccesible, or there's been no activity for 2 or 3+ years with no signs of life, and also the tool has not found any traction to create a community to maintain it ->  These are the main candidates to be removed. No point in keeping tools that don't work, as they can only create further confusion or display outdated methods for a given use-case. 

- **Tag the uncertain ones**: Some tools might just be in maintenance mode or the maintainers are busy. For these, if no consensus can be reached whether its active or not, a "Maintenance Status Unknown" tag can be potentially added so developers know what they're getting into and display them with lower priority.

## How?

To make all steps of the plan as transparent as possible, whats can be better than a public spreadsheet tracking the progress? This way, the process is less "trust me bro" and more of a transparent due diligence. Anyone can and should challenge the categorizations I make here, I'd rather be wrong and keep a good tool than unintentionally remove something that maybe still alive.

For some of the less obvious cases, we can have an open standing PR where a set duration time can be given to tool owners to reach out and correct the updated status.

In the worst case, a tool that was not inactive but actually a repository that does not need regular maintenance and fulfills all its stated purposes in 2025. In this case they can always reach out and reinstate themselves in the list with a new PR which also opens the chance for them to update some of their information in the portal.

## Preliminary Assesment

[...]

PR #1507, opened 2025-06-25
[Purge] Remove inactive tool: libada-go

0xBora:
## Summary

This PR removes the builder tool **libada-go** from the Developer Portal.

## Rationale

- Last commit made: +3 years ago
- A library maintained by an exchange Bitrue which has stopped maintaining the library since then. Reached out to @wolf31o2 and confirmed this is a suitable candidate for removal from the portal.

> Related Issue #1485


gufmar (approved):
Lgtm

PR #1491, opened 2025-06-25
[Purge] Update inactive tool: Fracada -> Fracada v2

0xBora:
## Summary

This PR updates the builder tool **Fracada** to **Fracada v2** the Developer Portal.

## Rationale

- Fracada v1 has been publically archived by dcSpark and deprecated (Unaudited). Repository now refers to Fracada v2. 

>## Notes
• Updated the preview image to match the format used in v1.
• Redirected links to the updated, audited repository.
• Removed the “favorite” tag, as Fracada v2 is also relatively old and no longer actively maintained.

PR #1514, opened 2025-06-27
[Purge] Remove inactive tool: ansible-cardano-node

0xBora:
## Summary

This PR **opens the discussion** to remove the builder tool **ansible-cardano-node** from the Developer Portal.

## Rationale

- Last updated 2 years ago and can be outdated today/not reflect best practices. This would make the playbook outdated and likely incompatible with current mainnet requirements/workflows No signs of ongoing maintenance or support for newer node versions. 

Reminder: This PR is intends to open the discussion/gather input to determine whether it is a tool used by any SPOs in 2025. A discussion before approvals/merge would be ideal.

> Related Issue #1485


rphair (approved):
As an SPO I can confirm the build process has a few things different in it today than 2+ years ago.  There are also other node usage questions today — especially whether to put the UTxO set on RAM or disk — that wouldn't have been configuration options 2 years ago.  I think it's to the point where it should be removed unless updated promptly.

I'll resolve next merge conflict by hand (not possible automatically: the occasional result of too many merges into the same file).  Bad luck this change got sucked into https://github.com/cardano-foundation/developer-portal/pull/1491 😬

rphair (approved):
Good fix @katomm @0xBora and I'll be watching for that more closely in the future 😬

PR #1518, opened 2025-06-27
[Purge] Remove inactive tool: piroutte

0xBora:
## Summary

This PR removes the builder tool **piroutte** from the Developer Portal.

## Rationale

- Last commit: +2 years ago.
- Has been deprecated and is not maintained anymore.

## Highlight

This repository might be one of the tools that deserve an "outdated" label which would not remove the tool entirely and keep it as an honorable mention. Such an idea was explored in a conversation found in the parent issue #1485 with @katomm. 

This PR is a good example to serve as a discussion point to justify a new "outdated" tag, as this tool it serves a unique enough domain. It can be argumented that keeping it in the developer portal can foster newer research initiatives based upon this. 

Or not.

> Related Issue #1485


rphair (approved):
Would also be happy to bump this over to an `outdated` tag if that structure is created.

katomm:
I agree that this would be the first good example for such an outdated tag. To ensure that it is not removed once it receives enough reviews, it is better to make the PR a draft for the time being.

Issue #1528, opened 2025-06-30
🔄 Portal 2025 - 1.2 - The Purge - Showcase


0xBora:
## Statement

Showcase section has many products listed, but some of them don't work. The rationale of this issue is explained under #1484, and had its first exploration in #1485.

## Proposal

Following the same systematic approach used for Builder Tools cleanup, I've checked each project and found dApps returning 404 errors, NFT projects with broken minting sites, wallets abandoned since 2021, and services that no longer exist.

## How?

Showcase items are a different story from builder tools. they are more about showing off what cool stuff’s been built on Cardano. As long as they’re still at least minimally functional, it is okay if they are a bit out of date or have alternatives that fulfill the same function "better". 

That said, this PR isn’t a free pass for dead links, a 404 is still a no-go.

## Preliminary Assesment

This [assessment sheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vSytSZX3teGx5GyCNjsB9ReAo-W9S0HGy71C6BxhGiUDu5Fxwf5qc7OvzPTrxP8kZbLB6VHx59mAHrY/pubhtml?gid=1572656486&single=true) provides the baseline data for categorization and providing notes for each item marked as Inactive. It reflects the status of products in the Showcase section based on the applied identification criteria for active, inactive, uncertain or update categories.

<p align="center">
  <img width="497" alt="Image" src="https://github.com/user-attachments/assets/e6d74332-e25b-406c-b83f-ee9cbf447586" />
</p>

> ### Quick Recap
> 14 products are labeled inactive, 1 uncertain and 1 to be updated out of 107 showcase entries.

Issue #1548, opened 2025-07-02
🔄 Portal 2025 – 1.3 – The Purge – Ecosystem Outreach


0xBora:
## Statement

While previous steps of the Purge (Builder Tools (#1485) and Showcase (#1528)) focused on cleaning up what’s currently listed, now its time to find answers to "What’s not listed that should be?".

Some of the most used dApps today aren’t visible in the portal because no one ever submitted them. In many cases, contributors don’t have time, forget, or don’t prioritize sending a PR to the Developer Portal. As a result, the portal can feel incomplete or outdated despite the ecosystem continuing to evolve rapidly.

## Proposal

This sub-issue focuses on active outreach to fill those gaps. Rather than waiting passively for contributions, this phase flips the approach:

- Identify tools/products that should be listed but aren’t.
- Submit PRs myself based on publicly available info and ask for review/approval.
- Track everything transparently via a shared spreadsheet.

## How?

As usual, I will maintain a public outreach [tracker](https://docs.google.com/spreadsheets/d/e/2PACX-1vShMNz6Y994idtBt6nUEbOcAc-tEffbe3y8iNkWmjSJ_6uNvRdleipem6WjTAHevAmD3kIVVxiybBlb/pubhtml?gid=0&single=true) here.

- Open up the list for the community to suggest additional items or self-report missing entries.

This lets everyone see which parts of the ecosystem still need to be better represented. Community members/maintainers are encouraged to suggest entries and nominate projects/tools they believe are underrepresented.

Lets increase the visibility into the ecosystem.

rphair:
@0xBora (cc @katomm @rdlrt) there is a complication, causing merge conflicts, from all these PRs adding the new entry at the end of the file.  After the first one, whose rebase could be done trivially (through the GitHub UI):
- https://github.com/cardano-foundation/developer-portal/pull/1577

... there is now no way for `git` to determine an ordering of the new entries without a manual merge conflict resolution specifying which entry should appear first... since these few dozen segments of new code have all been added in exactly the same place.

I had been staying ahead of all other merge conflicts through GitHub UI — especially those that only affected deletions happening in different places in the file, so only getting a conflict if the deletions were adjacent — but I think there could be a better way of doing this and that it might be done more easily in a local `git` branch or maybe through `gh` (GitHub CLI).

[...]

PR #1357, opened 2024-11-19
Add Orcfax

ross-spencer:
👋 Hello there! Welcome. Please follow the steps below to tell us about your contribution.

1. Please complete a Checklist
2. Fill in all sections of the template
3. Click "Create pull request"

## Checklist

 <-- Please fill the boxes with [x] before submitting a pull request --> 

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 
- [ x I have not committed any changes to `yarn.lock` (or have [removed these changes](https://developers.cardano.org/docs/portal-contribute/#i-committed-yarnlock-to-my-pr-branch-what-do-i-do)).

## Showcase addition

<-- Provide information for every bullet in the list below. The tags you select must match the tags in your changes to the showcase.js  -->

* Title: Orcfax
* Description: Orcfax is a decentralized oracle service designed to publish data about real world events to the Cardano blockchain. Orcfax data is made available to on-chain smart contracts in Cardano's eUTXO native format using the Orcfax Protocol.
* Website:   https://orcfax.io
* Source: null
* Tags:
  * `oracle`

PR #1267, opened 2024-05-23
Add Charli3 oracle to Showcase + category cleanup

katomm:
## Description

- add Oracle category
- add Charli3 Oracle
- fix category sorting

## Checklist

 <-- Please fill the boxes with [x] before submitting a pull request --> 

- [x] I have read the [How to Contribute](https://developers.cardano.org/docs/portal-contribute/).
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/data/showcases.js)
- [x] I have run `yarn build` after adding my changes **without getting any errors**. 

## Showcase addition

<-- Provide information for every bullet in the list below. The tags you select must match the tags in your changes to the showcase.js  -->

* Title: *Charli3*
* Description: *Charli3 is a decentralized Oracle solution on Cardano, built natively for the chain, producing fully auditable data records on Cardano ledger.*
* Website:   https://charli3.io
* Source: none
* Tags:
  * `oracle`
  * `metadata`

PR #1552, opened 2025-07-05
Purge/orcfax

0xBora:
## Summary

This PR updates the showcase item **Orcfax** on the Developer Portal.

## Rationale

- Oracles are displayed in showcase as products built on Cardano. (see Charli3)

> Related Issue #1485, #1528

### DIFF
--- a/src/data/builder-tools.js
+++ b/src/data/builder-tools.js
@@ -1005,15 +1005,6 @@ export const Showcases = [
     tags: ["c", "serialization", "sdk", "lowlevel"]
   },
   {
-    title: "Orcfax",
-    description:
-      "Orcfax is a decentralized oracle service designed to publish data about real world events to the Cardano blockchain. Orcfax data is made available to on-chain smart contracts in Cardano's eUTXO native format using the Orcfax Protocol.",
-    preview: require("./showcase/orcfax.png"),
-    website: "https://orcfax.io",
-    getstarted: "https://docs.orcfax.io/consume",
-    tags: ["oracle"],
-  },
-  {
     title: "Lucid Evolution",
     description:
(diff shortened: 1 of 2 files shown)


rphair:
that does seem to be consistent with this earlier PR establishing Oracles in the Showcase:
* https://github.com/cardano-foundation/developer-portal/pull/1267

So then with the only remaining `oracle` gone from builder tools (with this PR), that category would also be empty and should be "cleaned up" (as in the earlier PR) by also removing the emptied category.

If not, maybe @katomm could offer some rule about whether to classify oracles as Showcase items or Builder Tools, if both are acceptable (as they effectively have been since #1357).

rphair (commented):
@0xBora before I approve this in accord with https://github.com/cardano-foundation/developer-portal/pull/1552#pullrequestreview-3344383834, what do think about consequences of the Oracle category under Builder Tools (https://developers.cardano.org/tools?tags=oracle) being empty as soon as this item is moved?  (already asked in https://github.com/cardano-foundation/developer-portal/pull/1552#issuecomment-3040347995 but seems worth repeating now)

Specifically: should this PR not include a code change that removes the emptied category?  Or perhaps you have more items on the way to fill it?

0xBora:
@rphair you’re right, moving Orcfax to Showcase means we would now present oracle protocols as products rather than tools going forward. It also clarifies how we should categorize future oracle-related projects consistently. I’ve removed the “Oracles” category from the `builder-tools` list accordingly now.

This is also a good reminder that it’s time to revisit my earlier plan in #1585 to cleanup the list of filters we have at hand and come up with a modern list of filters that represents the projects more accurately. 

Since both the Showcase items and Builder Tools data are now in a stable state after the major cleanup and updates we did in #1485, #1528, and #1548 this seems to be a good time to start with that process.

PR #1580, opened 2025-07-06
Get started with Cardano CLI: Add era to stake-address

SimonSchubert:
I received the following error with version **10.11.1.0** on macos and it seems like the era was missing from the command
```
cardano-cli stake-address key-gen \
--verification-key-file stake.vkey \
--signing-key-file stake.skey
Invalid argument `stake-address'

Usage: cardano-cli 
                     ( address
                     | key
                     | node
                     | hash
                     | query
                     | legacy
                     | byron
                     | conway
                     | latest
                     | debug commands
                     | version
                     | cip-format
                     | compatible
                     )
```

### DIFF
--- a/docs/get-started/cardano-cli/get-started/get-started.md
+++ b/docs/get-started/cardano-cli/get-started/get-started.md
@@ -74,7 +74,7 @@ Testnet addresses start with 'addr_test' and mainnet addresses with 'addr'.
 ### Generate a stake key pair 
 
 ```shell
-cardano-cli stake-address key-gen \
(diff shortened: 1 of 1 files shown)


rdlrt:
Yes, the current status is a bit confusing (inconsistent) on cardano-cli 's root interface, as some commands were only available to root and others not - was raised [here](https://github.com/IntersectMBO/cardano-cli/issues/1015) where it is mentioned there is a change in PR (But cant locate the actual PR).

SimonSchubert:
understandable, thank you 👍

PR #1646, opened 2025-09-12
Update NEWM entry in Showcase

cristhianescobar:
👋 Hello there!
--
 Updating NEWM entry information 
 
Screenshot 2025-09-12 at 1 45 42 PM

- [x] I have read the [Contributing Guidelines](https://github.com/cardano-foundation/developer-portal/blob/staging/CON…
- [x] I have read the [Showcase Requirements](https://github.com/cardano-foundation/developer-portal/edit/staging/src/d…


rphair (commented):
@cristhianescobar thanks for stepping in to refurbish this old entry.

First, I believe — to properly review this — the array item for NEWM needs to be put back into its original position.  If you think your sequence in the array gives you some disadvantage — which would surprise me, since I thought the ordering on the site was established by other means — then please post here (@0xBora might also have some ideas about whether ordinality could be a problem for anyone).

After that's done (or not done, if there are confirmed grounds for moving the listing) I have a couple more editorial suggestions, but would prefer to wait until the issue above is addressed before posting these (to avoid having to split or duplicate the review comments).

cristhianescobar:
> @cristhianescobar thanks for stepping in to refurbish this old entry.
> 
> First, I believe — to properly review this — the array item for NEWM needs to be put back into its original position. If you think your sequence in the array gives you some disadvantage — which would surprise me, since I thought the ordering on the site was established by other means — then please post here (@0xBora might also have some ideas about whether ordinality could be a problem for anyone).
> 
> After that's done (or not done, if there are confirmed grounds for moving the listing) I have a couple more editorial suggestions, but would prefer to wait until the issue above is addressed before posting these (to avoid having to split or duplicate the review comments).

@rphair You're absolutely right! I got a bit too enthusiastic about NEWM and forgot about the established order 😅 

Moving it back to its proper place now. Bring on those editorial suggestions!

src/data/showcases.js
```diff
@@ -1106,11 +1106,11 @@ export const Showcases = [
   {
     title: "NEWM",
     description:
-      "Decentralized music streaming replaces middlemen with code that's not incentivized to act in favor of third parties that take a huge cut of streaming revenue.",
+      "At NEWM, we're on a mission to reshape the music industry by fostering direct, seamless connections between artists and fans.",
```
rphair, line 1109: I think the new `description:` is a bit better: although the initial `Decentralized music streaming` was a vital part of describing the service and should definitely remain there.  For the rest of the text, how about enumerating the features that are part of the decentralised offering, instead of (or perhaps in addition to) the marketing language; like:
* list the specific social features that "connect artists & fans"
* list the unique service features that "reshape the music industry"

Ideally, newcomers to the Cardano community should see a listing that makes sense here without knowing anything about Cardano, NEWM, or NEWM's history in the community.

rphair (commented):
that's all right @cristhianescobar, NEWM _is_ part of the "established order" 😅 i.e. I don't think anyone would dispute your application of the `favorite` tag... although FYI this tag is generally applied only by editors (I would myself be in favour of applying it, due to its longevity and influence).

src/data/showcases.js
cristhianescobar: Thanks for the feedback, let me get back with the team and we can finalize the description that makes the most sense.

rphair (approved):
thanks for https://github.com/cardano-foundation/developer-portal/pull/1646/commits/5361959f04419cfa2a0f39309c7176d6050f3e4f - looks great now.

src/data/showcases.js
rphair: Resolved with https://github.com/cardano-foundation/developer-portal/pull/1646/commits/5361959f04419cfa2a0f39309c7176d6050f3e4f.

rdlrt (approved):

PR #1789, opened 2026-05-12
Add matotam builder tool

MatoB13:
Adds matotam to the Cardano builder tools list.

Matotam is a Cardano wallet-to-wallet messaging tool where messages are minted and delivered as NFTs directly on-chain.

Features:
- wallet-to-wallet messaging
- no backend
- ADA Handle support
- optional encryption
- NFT-based delivery
- AI/agent compatible metadata


0xBora:
I think this proposed tool is quite early in its development cycle with a v0.1 beta tag attached to it. I dont think its a good candidate at the time to be included. Reason is fit with what Builder Tools is for rather than the project itself.

Builder Tools is a curated list of developer infrastructure: the SDKs, providers, indexers, and libraries that other developers build with and compose into their own apps and services. This seems to be an end-user application built on Cardano. It's closed source, and at its current stage it doesn't expose an SDK or API that other developers could integrate to add on-chain messaging to their own apps, so there isn't enough signal yet for it to sit in this list as developer infrastructure.

Glad to revisit in future. In the meantime an ecosystem apps showcase, like the apps directory on cardano.org, can be the better home for featuring it. Closing for now.

PR #1810, opened 2026-06-15
Portal 2026: Disrupt all and introduce the curriculum 

0xBora:
This branch reshapes the developer documentation from a categorized index into a learning path. The developer docs now read as a numbered curriculum, each concept has one authoritative page with the SDKs side by side as tabs, and the file tree finally has a single rule for where things live.

It is a large branch because most of the work is consolidation. The same material existed three and four times over, once per tool, and this collapses it.

## The problem

Portal 2026 (#1758) names the symptom: the 2025 cleanup bundled things together, but the portal kept growing into a categorized index rather than a cohesive learning experience. The same concept was explained once in each tool's flavor. Governance spanned three directories. Some content was organized by tool, some by task, some by language, so a contributor with a new page had no obvious home for it. And "Get Started" had become the place everything went, so the actual starting path was gone, which is exactly the point @CarlosLopezDeLara raised in #1639.

## What changed

**A seven-module curriculum (#1773).** Start Here, then Fundamentals through Production, then Integrations, Community, and Contributing. An ordered path from "what is Cardano" to "I shipped a dApp", which is the zero-to-MVP-on-testnet bar Portal 2026 anchors to. Start Here is a front door now, not a dumping ground.

**One page per concept, tools as tabs (#1772).** Staking, governance, wallet authentication, transaction building and the rest are single concept pages with Evolution, Mesh, and cardano-cli side by side, instead of a page per tool. This is the tool-agnostic structure @rober-m argues for in #1738: the concept is explained once, and removing or adding a library does not change the explanation.

**A legible file structure (#1809).** The tree is audience-first: developers, operators, community, contribute. The cardano-cli silo, the per-language pages, and the standalone governance section dissolve into the concept pages or link out. One rule holds: content lives in the module or audience where it is used. That split is also the groundwork for the wider taxonomy question @katomm raises in #1292, without taking a position on the rebrand itself.

**LLM-native groundwork (#1830).** An AI-assisted development page points developers at current, authoritative context for their coding agents, and the overview separates AI as a build aid from autonomous on-chain agents.

## One honest divergence

@CarlosLopezDeLara, this branch is SDK-first, so cardano-node and cardano-cli live in Production rather than at the front. The reasoning is the zero-to-dApp onboarding bar, but your point about the node being the most direct route to the fundamentals is fair, and I would like your read on whether a node-first entry belongs earlier.

[...]

Issue #1758, opened 2026-04-13
Portal 2026


0xBora:
Last year, with [Portal 2025](https://github.com/cardano-foundation/developer-portal/issues/1484), we ran the first structured pass over a plan: the project showcase got moved, builder tools got a slight refresh, and the fundamentals were rearchitected reasonably to get started faster. That effort wrapped up earlier this year.

## What comes next?

Portal 2026 pulls together a scope of portal-relevant work drawn from several sources: feedback from community developers, the feedback loops of existing developer onboarding efforts, discussions Intersect Developer Experience working group, and opinions surfaced across broader ecosystem conversations. It's not a top-down mandate and aims to clearly strategize a coalescence of all ideas to improve the portal.

The cleanup worked, but the portal still has the symptoms of organic growth beyond what the first pass could address. The same concept gets explained three times in tool-flavored places. Governance content spans three directories with overlapping titles. 

So Portal 2026 is less about removing debt and more about building on top of the cleaner base. The Developer Portal should be the canonical learning platform for Cardano developers: up-to-date, highly curated, and designed to lower the cost of acquiring new developers for the ecosystem. The work that gets us there is creating standards and canonical patterns for how Cardano development is explained, building a baseline understanding of developer onboarding through curated content, and continuously exploring what the portal can become so it best serves the ecosystem over time. more developers feed into hackathons, build teams, and ship things onchain. More developers building means more innovation and more transactions moving through Cardano. That's the bigger reason any of this matters.

The concrete bar/KPI we can anchor to is to ensure we can equip developers being onboarded to get them from **zero to MVP on testnet in the shortert possible time, using the portal alone**. Everything in the pillars below is there to move new developers closer to that, and to evaluate the existing information on the portal to be shaped building upon this idea.

Portal 2026 is written as an open-ended strategy rather than a finalized checklist. The pillars below define directions the portal should move in, and the sub-issues under each are where contributors can propose and track specific work as it materializes.

I've structured the scope into six pillars. Unlike last year's sequential steps, these run in parallel. Two of them (cleanup and the portal structure work) act as prerequisites that benefit the others, so they can get priority when possible.

### Pillar 1: One story with many differnt tools
> **Sub-issues:** [Pillar 1](https://github.com/cardano-foundation/developer-portal/issues/1772)

[...]

PR #1820, opened 2026-06-24
Reorganize the contract library by use case

0xBora:
The contract library page is a MeshJS-only table that never surfaced the CF cardano-template-and-ecosystem-monitoring use-cases. This reorganizes it into a use-case index: each pattern (escrow, vesting, swap, payment splitter...) lists the best sources, including the ecosystem-monitoring multi-language reference (Aiken on-chain + MeshJS / Evolution / PyCardano / CCL off-chain), plus use cases that were missing (auction, crowdfund, HTLC, vault, lottery).

It stays a curated index, not another browser; the portal points at the sources rather than hosting contracts, and cross-links the templates gallery (#1816), the "start an app" sibling to this "find a contract" page.

Stacked on #1810. Part of Portal 2026 (#1758).

### DIFF
--- a/docs/developers/curriculum/05-smart-contracts/contracts-library.md
+++ b/docs/developers/curriculum/05-smart-contracts/contracts-library.md
@@ -2,21 +2,35 @@
 id: contracts-library
 title: Contract Library
 sidebar_label: Contract library
-description: A catalog of open-source Cardano smart contracts with live demos and end-to-end source, from Hello World to escrow and marketplaces.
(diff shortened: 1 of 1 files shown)


0xBora:
Superseded by the Contracts tab in #1816, which surfaces the full use-case index (all 21 ecosystem-monitoring use cases, each with its on-chain/off-chain implementations, plus the MeshJS and Anastasia ones) at /templates/contracts. The contract-library page now points to that browser. Closing.

PR #1888, opened 2026-07-12
Air Gap Environment: correct number of options

rphair:
Very basic change: number of options presented was not increased from 2 to 3 after adding Nix `cardano-airgap`.

Could be improved further — technically there more options, maybe even an infinite number — though I think keeping the language as simple as it is now is a better introduction for the material that follows.


0xBora (approved):
Thank you @rphair, merging this straight in!

PR #1905, opened 2026-07-23
Add extractRegion helper for embedding tested example snippets

### LINKED ISSUE #1904: Import snippets from real examples
### Why

We need to have trustful code snippets that can be checked if they still works. It avoids that the snippets stop working without us knowing.

### What

To have code snippets that works and can be checked by CI periodically, we need to import from real project examples that can be tested and the CI can execute, so if the example is not valid anymore we can get a notification from the CI that the example needs to be updated/fixed.

### How

docusaurus supports the `raw-loader` to load a file, but it wouldn't be enough because it would show the whole file in the code snippet box, so a custom `extract region` needs to be created and a convention need to be defined to extract only a piece of code in a file. 

The convention:
We add a comment to start the region with an id and a comment to close the region, for example

```js
// #region any-id

// we can add comments too
const variable = "variable"

// #endregion any-id
```  

so in the mdx file, we load the file and use the custom extract region function
```mdx
import extractRegion from "@site/src/utils/extractRegion";
import ExampleFile from "!!raw-loader!@site/FOLDER/example.js";

<CodeBlock language="javascript" title="examples.js: a description about this snippet">
  {extractRegion(ExampleFile, "any-id")}
</CodeBlock>

```

paulobressan:
Close #1904. From that issue:

### Why

We need to have trustful code snippets that can be checked if they still works. It avoids that the snippets stop working without us knowing.

### What

To have code snippets that works and can be checked by CI periodically, we need to import from real project examples that can be tested and the CI can execute, so if the example is not valid anymore we can get a notification from the CI that the example needs to be updated/fixed.

### How

docusaurus supports the `raw-loader` to load a file, but it wouldn't be enough because it would show the whole file in the code snippet box, so a custom `extract region` needs to be created and a convention need to be defined to extract only a piece of code in a file. 

The convention:
We add a comment to start the region with an id and a comment to close the region, for example

```js
// #region any-id

// we can add comments too
const variable = "variable"

// #endregion any-id
```  

so in the mdx file, we load the file and use the custom extract region function
```mdx
import extractRegion from "@site/src/utils/extractRegion";
import ExampleFile from "!!raw-loader!@site/FOLDER/example.js";

<CodeBlock language="javascript" title="examples.js: a description about this snippet">
  {extractRegion(ExampleFile, "any-id")}
</CodeBlock>

```


0xBora:
Imo a good direction.

A few things on the helper itself:

1) Small one in the CONTRIBUTING text: "pull that region into an .mdx page" would make this the repos first .mdx file. Everything under docs/ is .md, imports already work there, and all pages use them.

2) The bigger one is `docusaurus-plugin-llms`, which generates llms.txt and the per-page .md twin behind the "Copy page as markdown" button. It doesn't seem to resolve raw-loader imports. I ran our current config against a page written this way: the import lines get stripped and <CodeBlock language="ts">{extractRegion(Source, "build")}</CodeBlock> is passed through as literal text, so the code never appears at all. A pasted fence shows up in those outputs fine, and we'd want an imported snippet to as well. That surface is what AI tooling consumes, so it's the one place this could quietly cost us something.

paulobressan:
Thanks for your review. I haven't noticed both conditions. I'm changing the CONTRIBUTING file and verifying the plugin to support the imports

paulobressan:
Both points are fixed. CONTRIBUTING now says .md instead of .mdx.

On the llms output you were right. raw-loader is a webpack feature, and docusaurus-plugin-llms never goes through webpack, it reads the source markdown from disk, strips the imports and copies the <CodeBlock> out as literal text, so the code never reaches llms.txt or the "Copy page as markdown" files. The fix is a post-build script, scripts/fix-llms-snippets.js: the imports are gone from the generated output but still in the source doc, so it reads them back to find each file, extracts the region with the same `extractRegion` the page uses. 

@0xBora @rober-m

rober-m (commented):
I tested it following the instructions, and it worked correctly. 

One thing we might mention in the instructions is that when running `yarn start`, the contents of the "copy page" button copy the website's code, not the content. It's not an issue since `yarn build && yarn serve` works, but it may confuse new contributors.

rober-m (approved):

Issue #1839, opened 2026-06-30
Portal 2026 - Pillar 5: Periphery and upkeep


0xBora:
Umbrella for portal periphery and upkeep: the non-content plumbing that keeps the repo and the deployed site clean and correct.

Scope: CI workflows, redirects, security headers and CSP, PR and issue templates, README and CONTRIBUTING, site metadata, and search config.

These are passthrough fixes that do not belong under the content pillars (1 to 4). Individual fixes are tracked as sub-issues here.

Part of #1758.

PR #1930, opened 2026-07-30
Restructure the smart contracts module into classes and a reference shelf

0xBora:
Stacked on #1929, which moves the security subtree out of `advanced/`. Review that one first; this branch contains its commit.

Module 5's security section was a filing cabinet. Nineteen vulnerability pages sat behind a flat nineteen-row table, most of them under thirty lines and sharing one rigid template, and eleven were reachable only from that table. Reading them taught you nothing about how the failures relate, because the structure said they were all equally distinct.

They now group into four classes, each page opening with what unites its members: resource exhaustion, unchecked inputs, staking and certificates, and evaluation and grinding. The four genuinely deep pages stay standalone. Every folded body is preserved byte for byte, so the writing is unchanged; only the intros and the reference page are new.

Two stubs dissolve into the security page's verification section, which is where a reader looking for them already was. Alongside that, the Advanced category gains an overview so it stops being a bare folder, and `optimization.md`'s nineteen flat headings group under five named sections following the page's own summary themes.

Related to #1839


0xBora:
Superseded by #1931, which was stacked on this branch and therefore carried all of these commits into staging when it merged. Every change from this PR is live on staging: the four vulnerability class pages, the audits and formal-verification merge, the advanced overview, and the optimization regrouping are all present, and the dissolved pages are gone.

Closing rather than merging because the branch now conflicts with staging and re-applying it would revert the Module 6 work that merged alongside it.

PR #1932, opened 2026-07-30
Fix contradictory facts, unbacked claims, and affiliate links across the portal

0xBora:
The recent restructure fixed how the curriculum is organized. It never checked whether the content underneath was correct, and an audit says it often was not.

Facts contradicted each other across pages. `k` meant two different numbers depending on where you read it, and the page that declares itself the canonical min-ADA reference disagreed with its own code examples by enough that a reader following them would overspend by 67%. Each of those now states one number with one meaning. A security property on the DeFi page was written as a tendency rather than the requirement it is, and the last era-pinned `cardano-cli conway` invocations join the sixty already using `latest`.

Numbers that rot are now marked as such: protocol-governed values name their parameter and say "currently", following the pattern the governance page already set by querying its deposit live. And the wallet page called a template's configuration "build-verified" when nothing builds it. Rather than adding a job to make the word true, the claim is gone, because neither template has a lockfile and a build check would measure whether the latest of forty upstream packages works today rather than whether the committed template works.

Hardware sourcing becomes one vendor-neutral surface. Six IoT pages carried eighteen affiliate tracking redirects; the hardware reference now lists each part by the search term a supplier uses, names the substitutes that work, and the workshops point there instead of keeping their own lists.

Two things left for you rather than decided here. The recommended Mesh starter pins a beta release, which is a judgment call about how closely templates track upstream. And `docs/operators/` has the same era-pinning issue, untouched because pinning may be deliberate for operators on a fixed node version.

Related to #1839


0xBora:
Added a fifth commit after a repo-wide sweep found affiliate tracking outside the curriculum too.

`blog/2024-02-19-february.md`, the Rejuve.AI interview, carried five affiliate-coded links across four networks, inside the interviewee's answer listing partner vendors:

- `avea-life.com/?ref=RejuveAi`
- `peptide-bioregulator.com/?rfsn=7663179.dd70b5`
- `trudiagnostic.com/?rfsn=...&utm_medium=affiliate&...`
- `apolloneuroscience.pxf.io/...` (Impact)
- `travala.sjv.io/Rejuve` (Sovrn)

The first three keep their destination with the referral parameter removed, so the quote reads identically and the reader still reaches the vendor. The last two are affiliate redirect domains with no clean destination to preserve, so those two vendors are named in plain text instead of linked. No wording in the interview changed.

Worth noting these are health and longevity supplement vendors, which is a category where a monetized redirect from a Foundation domain reads worse than most.

PR #1985, opened 2026-08-24
Refresh vulnerable transitive dependencies and drop dead ones

0xBora:
Dependabot flags 23 alerts on the site lockfile, all in transitive build-time dependencies. Its own fix attempts fail on our yarn 1 lockfile, so this refreshes the resolutions manually: brace-expansion, js-yaml, ws and ajv move to their patched versions within the ranges our dependencies already declare. That resolves 17 of the 23 alerts, including every high severity one that has a compatible fix.

The remaining six are documented rather than forced: image-size has no patched release yet, and serialize-javascript, qs and uuid sit behind major versions their parents do not accept. This also drops the unused node-fetch dependency (nothing imports it, Node has fetch built in) and the dead deploy script (Netlify handles deploys), and adds a comment explaining the sharp pin from #1883.

Related to #1839

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "start": "yarn build-stats && yarn build-og && docusaurus start",
     "build": "yarn build-stats && yarn build-og && docusaurus build && node scripts/fix-llms-paths.js",
     "swizzle": "docusaurus swizzle",
-    "deploy": "docusaurus deploy",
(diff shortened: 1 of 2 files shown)


0xBora:
Local/CI build both pass, and fixed versions are aligned with Githubs advisorial data

PR #1988, opened 2026-08-25
Enhance SPO voting process documentation to clarify multiple votes in a single transaction

thenic95:
Thanks for contributing. Complete the checklist, fill in the section below, and create the pull request.

## Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/cardano-foundation/developer-portal/blob/staging/CONTRIBUTING.md).
- [x] I have not committed any changes to `yarn.lock` (or have [removed these changes](https://github.com/cardano-foundation/developer-portal/blob/staging/CONTRIBUTING.md#faq)).

The current SPO Governance Page does not indicate that the pools can vote on more than one governance action in a single transaction. I addressed this by adding examples and showcasing how this can be done.


docs/operators/governance/spo-governance.md
```diff
@@ -148,7 +181,18 @@ cardano-cli conway query proposals \
   | jq '.[0].stakePoolVotes'
 ```
 
-Your pool ID should appear in the `stakePoolVotes` object with your chosen vote.
+Your pool ID should appear in the `stakePoolVotes` object with your chosen vote (`VoteYes`, `VoteNo` or `Abstain`). If you voted on several actions in one transaction, check each action ID in turn — or list every vote your pool has cast in one pass:
```
katomm: `query proposals --all-proposals` only returns proposals still live in the ledger. Once an action is ratified, enacted or expired it drops out of the output, so its vote no longer shows up here (and proposals submitted in the current epoch are excluded as well). Scoping the claim to active proposals avoids operators concluding a vote was never recorded:

```suggestion
Your pool ID should appear in the `stakePoolVotes` object with your chosen vote (`VoteYes`, `VoteNo` or `Abstain`). If you voted on several actions in one transaction, check each action ID in turn — or list your pool's votes on all currently active proposals in one pass:
```

docs/operators/governance/spo-governance.md
```diff
@@ -148,7 +181,18 @@ cardano-cli conway query proposals \
...
+
+```shell
+# stakePoolVotes is keyed by the hex-encoded pool ID, not the bech32 pool1... form
+POOL_ID=$(cardano-cli conway stake-pool id \
+  --cold-verification-key-file cold.vkey --output-hex)
```
katomm, line 189: `--output-hex` only exists since cardano-cli 10.9.0.0. Older releases still in use only accept `--output-format hex` (deprecated but working everywhere), which is also the spelling [register-stake-pool.md](https://github.com/cardano-foundation/developer-portal/blob/staging/docs/operators/block-producer/register-stake-pool.md#L193) uses for the same command. A short comment in the snippet would cover operators on older node bundles, for example: `# on cardano-cli < 10.9.0.0 use --output-format hex instead`.

docs/operators/governance/spo-governance.md
```diff
@@ -206,6 +250,7 @@ See [Calidus Keys](../../operator-tools/calidus-keys) for setup instructions.
 ## Key points
 
 - **One vote per proposal per pool.** Submitting a second vote overwrites the first.
+- **Many proposals per transaction.** One vote file per governance action, all of them in a single transaction, signed once with your cold key. Use it when several actions are live so your cold key comes out of the air gap once instead of once per proposal.
```
katomm: "your cold key comes out of the air gap" reads as if the key itself leaves the air-gapped machine, which the security section on this page explicitly rules out (only the transaction files travel). The Step 1 tip already phrases it safely with "unlock":

```suggestion
- **Many proposals per transaction.** One vote file per governance action, all of them in a single transaction, signed once with your cold key. Use it when several actions are live so you unlock your cold key once instead of once per proposal.
```

katomm (commented):
Thanks, the batching guidance is a great addition. The CLI mechanics all check out (repeatable `--vote-file`, `--witness-override 2` staying correct, the vote merging conflict on duplicate votes, hex-keyed `stakePoolVotes`). Three small points inline.

0xBora:
Thanks for the contribution. @katomm made comments that make some of the nuances clearer. Happy to merge once those are in.

thenic95:
Thank you @katomm 

I accepted all of the suggestions and made the suggested changes @0xBora

PR #1991, opened 2026-08-28
Rename drep-always-abstain to always-abstain

Godspeed-exe:
Updating to match cardano-cli v11.0.0

This repository uses PR templates. Click the `Preview` tab above and pick the one that fits your change:

- [Add Builder Tool](?expand=1&template=builder-tool.md)
- [Update documentation or Fix a Bug](?expand=1&template=standard-change.md)

(The Preview tab renders these as clickable links; selecting one reloads the form with the right template.)


0xBora (approved):
Thanks, good catch. Retargeted to staging, which is the branch we merge into (main only mirrors it for production)

PR #2003, opened 2026-09-08
Onboarding: Intermediate lectures 10 to 13, with vesting, gift card and oracle examples

paulobressan:
Adds the four design lectures that close the Intermediate track, each with a runnable Aiken project and, where the lecture has an app, a Mesh app.

## Lectures

- **10, Handling time (vesting)**: a deadline enforced through the transaction's validity window, without the contract reading a clock.
- **11, Multi validators (a gift card)**: one script with a `mint` and a `spend` handler, and a one-shot policy seeded from a UTxO.
- **12, Modifying state (an oracle)**: a rate that stays on the chain and changes. A beacon policy minted once, and an oracle whose `Update` puts the UTxO back and whose `Delete` burns the beacon. 
- **13, Reference inputs & reference scripts**: a consumer that reads the oracle's datum as a reference input, then the same consumer published once as a reference script and unlocked through it.


docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+The design in one sentence: **the funds go to the person named in the datum, and only in a transaction that happens after the date in the datum.**
+
+A contract cannot read a clock, because a clock gives a different answer every time you ask it. **[On-chain vs off-chain](/docs/developers/onboarding/lectures/intermediate/on-chain-vs-off-chain#why-the-split-exists)** explained why.
+
+## The window, not the moment
+
+The solution is the one you met in Beginner, in [Time on Cardano](/docs/developers/onboarding/lectures/beginner/time-on-cardano). Every transaction can carry a **validity window**, and **you** set it when you build the transaction. It is a statement the transaction makes about itself: **this transaction may only be included in a block _after_ this slot, and _before_ that slot.**
+
+The window has two ends, and each has a name you will meet in code: the **lower bound** (`invalid_before`), and the **upper bound** (`invalid_hereafter`, also called the TTL, for time to live). A deadline is a rule about being late, so this contract needs the lower bound.
```
rober-m: > A deadline is a rule about being late, so this contract needs the lower bound.

It doesn't seem obvious from what the developer knows at this stage.

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+- An investor who cannot sell for six months after a token sale.
+- A seller who is paid only once the buyer's refund window has closed.
+- A researcher whose grant arrives in four instalments over two years.
+
+**Vesting** is the name for it.
+
+The company must not be able to change its mind on the day before the date. An app cannot enforce that, because the company controls the app.
+
+## From idea to contract
```
rober-m:
```suggestion
## From idea to architecture
```

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+    Led->>Led: it is now 13:30. Is that<br/>inside 13:00 to 15:00?
+    Led->>Val: yes, so run the contract
+    Val->>Val: does 13:00 come after 12:00?
+    Val-->>You: yes, the funds move where the claim says
+```
+
+- The **ledger** asks: is the current slot inside the window the transaction declared? A transaction outside its own window is rejected, and no contract runs at all.
+- The **validator** asks: does that window start after the deadline in the datum? The ledger cannot ask this one for you. A deadline is one contract's rule, written in one datum, and the ledger does not read datums.
+
+Remove the validator's check and anybody could claim on the first day with a perfectly honest window. Remove the ledger's check and the window becomes a claim that nobody verified.
```
rober-m: Confusing. Someone could think: "Can I remove ledger rules?"

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+import TabItem from "@theme/TabItem";
+import CodeBlock from "@theme/CodeBlock";
+import extractRegion from "@site/src/utils/extractRegion";
+import VestingAiken from "!!raw-loader!@site/examples/onboarding/lectures/intermediate/vesting/on-chain/aiken/validators/vesting.ak";
+
+# Handling time: vesting
+
+Someone describes an idea to you. Your job is to decide what the contract has to remember, which actions it has to allow, and what it has to refuse. That decision is the **design**.
+
+The design here starts from the rule: **not before a certain date**.
```
rober-m: I wouldn't say this. Both checks (beneficiary and deadline) are equally important.
```suggestion
```

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+import CodeBlock from "@theme/CodeBlock";
+import extractRegion from "@site/src/utils/extractRegion";
+import VestingAiken from "!!raw-loader!@site/examples/onboarding/lectures/intermediate/vesting/on-chain/aiken/validators/vesting.ak";
+
+# Handling time: vesting
+
+Someone describes an idea to you. Your job is to decide what the contract has to remember, which actions it has to allow, and what it has to refuse. That decision is the **design**.
```
rober-m:
```suggestion
Let's start approaching our work like a real-world scenario. Someone (co-founder, investor, boss) describes an idea to you. Your job is to decide what the contract must remember (state), what actions it must allow (state transitions), and what it must refuse (checks) to meet the product requirements. Those decisions will allow you to **design** your contract/protocol.

Let's see today's requirements.
```

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+The design here starts from the rule: **not before a certain date**.
+
+## The idea
+
+A company hires a developer and promises them tokens. The tokens belong to the developer, but not today. The developer can take them after one year.
```
rober-m:
```suggestion
A company hires a developer and promises them tokens. The tokens belong to the developer, but they can't use them because they are locked. The company will vest them in four installments (once per year) over four years.
```

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Vesting** is the name for it.
+
+The company must not be able to change its mind on the day before the date. An app cannot enforce that, because the company controls the app.
+
+## From idea to contract
+
+Four questions turn an idea into a contract. Ask them in this order.
+
+**1. What has to be remembered?** This becomes the **datum**. A vesting contract has to remember two things: **who** may take the money, and **when** they can take it. Nothing else. The amount does not need to be remembered, because the UTxO already holds it.
```
rober-m: For a simple validator (we'll cover more steps in advanced lectures):

1. Create requirements as the first step
2. What actions are possible (allowed state transitions)
3. What must be true for each action? (what do we check?)
4. What has to be remembered to be able to run the checks?

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+A contract cannot read a clock, because a clock gives a different answer every time you ask it. **[On-chain vs off-chain](/docs/developers/onboarding/lectures/intermediate/on-chain-vs-off-chain#why-the-split-exists)** explained why.
+
+## The window, not the moment
+
+The solution is the one you met in Beginner, in [Time on Cardano](/docs/developers/onboarding/lectures/beginner/time-on-cardano). Every transaction can carry a **validity window**, and **you** set it when you build the transaction. It is a statement the transaction makes about itself: **this transaction may only be included in a block _after_ this slot, and _before_ that slot.**
+
+The window has two ends, and each has a name you will meet in code: the **lower bound** (`invalid_before`), and the **upper bound** (`invalid_hereafter`, also called the TTL, for time to live). A deadline is a rule about being late, so this contract needs the lower bound.
+
+The ledger and the validator both check that window, and they ask different questions. The examples use clock times, which are easier to read than slot numbers:
```
rober-m: - The ledger checks if the current slot is within the lower and upper bounds.
- The ledger transforms slots to POSIX.
- The validator has access to these bounds **and it can trust that the tx is happening within the bounds**.

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+    style I2 stroke-dasharray:4 3
+```
+
+The deadline travels between the two transactions inside the datum. The lock writes it, and the claim is the only thing that reads it. The claim's window is not in the datum. It comes from the second transaction, so the contract has both and can compare them.
+
+The app has a real clock, so the app turns your deadline into a **slot number** and writes that slot into the transaction.
+
+The validator never sees that slot. Once the ledger has checked the window, it converts the window into **POSIX milliseconds**, which is the number of milliseconds since 1 January 1970. Only then does it run the script. So `lock_until` in the datum is a date.
+
+Slot length is a network parameter, so a hard fork could change it. A rule written in slot numbers would then mean a different moment, and nothing would warn you. A date always means the same moment.
```
rober-m:
```suggestion
Slot length is a network parameter, so a hard fork could change it. A check written in slot numbers would then refer to a different moment, and the beneficiary could potentially consume the UTxO before the desired deadline. That's why we use POSIX, so the date always means the same moment in time.
```

docs/developers/onboarding/lectures/intermediate/10-handling-time.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Why the claim must declare a window
+
+Both bounds of the window are optional. A bound you leave out is treated as **infinite**. If the transaction declares no lower bound, it says "I have been valid since the beginning of time". That statement proves nothing about a deadline, so the check refuses the transaction.
+
+:::warning A deadline far in the future is an estimate
+Converting a date to a slot meets that same network parameter, from the other side. When an SDK converts a date, it assumes that slots keep the length they have today. As [Time on Cardano](/docs/developers/onboarding/lectures/beginner/time-on-cardano) explained, the conversion is only reliable a fixed distance ahead, currently about a day and a half (36 hours).
+
+The real risk is that you get no warning. If you ask an SDK to convert a date five years from now, it returns a slot number and reports no error. It only calculates with today's parameters. Our example locks funds for two minutes, which is safely inside the reliable range.
```
rober-m: Correct, but the node rejects the transaction in this case.

rober-m (changes requested):
Also, add only the key parts of the off-chain.

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+import TabItem from "@theme/TabItem";
+import CodeBlock from "@theme/CodeBlock";
+import extractRegion from "@site/src/utils/extractRegion";
+import GiftcardAiken from "!!raw-loader!@site/examples/onboarding/lectures/intermediate/giftcard/on-chain/aiken/validators/giftcard.ak";
+import validatorHash from "@site/src/utils/validatorHash";
+import GiftcardBlueprint from "@site/examples/onboarding/lectures/intermediate/giftcard/on-chain/aiken/plutus.json";
+
+# Multi validators: a gift card
+
+The last lecture had one action and one rule. This contract needs two things you have not met yet: **state that is not in a datum**, and **one script that guards two different actions**.
```
rober-m:
```suggestion
The last lecture had one purpose and one rule. This contract needs two things you have not met yet: **a state that is not in a datum** and **one script that guards two different purposes**.
```

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+The card is a **physical object**, and the shop keeps no list of who has one. The shop destroys the card when the customer hands it over.
+
+A contract has to reproduce three things:
+
+- Anybody may hold the card, and holding it is enough.
+- The funds and the card must move together. The shop must not be able to release the goods and keep the card, and the customer must not be able to keep the card and take the goods.
+- There is one card. The shop must not be able to print a second one for the same funds.
+
+## From idea to contract
```
rober-m: this is using the old steps

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+The same four questions as the last lecture.
+
+**1. What has to be remembered?** Nothing at all in the datum. The last lecture wrote the beneficiary into the datum, because the contract had to know **who** may claim. A gift card has no named owner.
+
+So the state lives somewhere else: in the **token itself**. A token that exists means an unused card. A token that has been destroyed means a card that has been used. A token is a piece of state that anybody can hold and transfer, and it needs no datum.
+
+**2. What actions are possible?** Three of them:
```
rober-m: - Destroying and releasing funds is always a single action.
- What about transactions? We should be able to transfer the giftcard to anyone we want. Even if that's handled by the ledger, not our contract, it's still a requirement.

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+import validatorHash from "@site/src/utils/validatorHash";
+import GiftcardBlueprint from "@site/examples/onboarding/lectures/intermediate/giftcard/on-chain/aiken/plutus.json";
+
+# Multi validators: a gift card
+
+The last lecture had one action and one rule. This contract needs two things you have not met yet: **state that is not in a datum**, and **one script that guards two different actions**.
+
+## The idea
+
+A shop sells gift cards. You pay 50 ADA today and you get a card. You give the card to a friend. Your friend walks into the shop, hands over the card, and takes 50 ADA of goods. The card is now used up. Nobody can use it a second time.
```
rober-m:
```suggestion
Let's say there's a shop that sells gift cards. You go there and buy a gift card redeemable for 50 ADA. You give the card to a friend. Your friend walks into the shop, hands over the card, gets 50 ADA transferred into his account, and the card is destroyed. Nobody can use it again.
```

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+The card is a **physical object**, and the shop keeps no list of who has one. The shop destroys the card when the customer hands it over.
```
rober-m:
```suggestion
In this case, we're trusting the shop will transfer the ADA to our friend. But what if we don't trust them? In that case, we don't need to go to the store. We can encode the logic for this use case in a smart contract that ensures these rules are followed, and that our friend gets paid!
```

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
rober-m: Also, remember what we discussed on the call. We trust the issuer to lock the ADA in the utxo in this case. That's why we don't check the value locked.

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+**4. What breaks if a rule is missing?** Remove the check in the spend handler and the funds leave without any card being destroyed. Anybody can take the 50 ADA, with or without a card, and your friend is left holding a card that opens nothing. Remove the limit on creating and anybody can make a second card, destroy that one instead, and take the 50 ADA that belongs to the first.
+
+Let the card be destroyed on its own and your friend can throw it away without taking the goods. Only one card can ever exist, so nothing is left that can release the funds, and the 50 ADA stays where it is for ever. Nobody gains anything, which is what makes it easy to do by accident.
+
+The design in one sentence: **a card can only ever be created once, and the card and the funds have to move together in both directions: the funds are released only if a card is destroyed, and a card is destroyed only if the funds are released.**
+
+## What makes it an NFT
```
rober-m, line 44: Good explanation!

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+Beginner's NFT used a native script with a deadline in it. Once the deadline passes, nobody can mint under that policy again. Before it passes, whoever holds the key can still mint more, so the token is only certainly unique after the deadline.
+
+A validator can read the transaction, so it can ask for something stronger: **a specific UTxO must be spent in the transaction that creates the token**. A UTxO is consumed exactly once in the whole history of the chain. So a policy that asks for one can succeed at most once, and the card is unique from the block that created it.
+
+That UTxO is called the **seed**, and a policy built this way is called a **one-shot policy**. The seed comes from your own wallet, and it is an ordinary UTxO holding ordinary ADA.
+
+Creating a card is one transaction:
+
+```mermaid
```
rober-m, line 56: It's a bit confusing to read the contents because there are no visual indications that each line is a different/new phrase. There are no termination dots, nor do they start with capital letters or `-`. I think we should somehow address this.

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Two purposes, one hash
+
+Remember from **[validator purposes](/docs/developers/onboarding/lectures/intermediate/validator-purposes)** that one script has one hash, and that this hash acts as both its **address** and its **policy ID**:
+
+```mermaid
+flowchart TD
+    S["one script,<br/>one hash"] -->|as a policy ID| M["`**mint** handler
+    guards making and destroying cards`"]
+    S -->|as an address| P["`**spend** handler
```
rober-m:
```suggestion
    S -->|as the address payment key| P["`**spend** handler
```

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+    style I1 stroke-dasharray:4 3
+    style I2 stroke-dasharray:4 3
+```
+
+The spend rule asks whether a card was destroyed. It does not ask how many UTxOs that one destroyed card is releasing. Here each card guards a single locked UTxO, so the simpler question is enough. A contract that keeps several UTxOs at one address has to count them, and the Advanced track covers that problem under the name **double satisfaction**.
+
+## Two purposes, one hash
+
+Remember from **[validator purposes](/docs/developers/onboarding/lectures/intermediate/validator-purposes)** that one script has one hash, and that this hash acts as both its **address** and its **policy ID**:
```
rober-m:
```suggestion
Remember from **[validator purposes](/docs/developers/onboarding/lectures/intermediate/validator-purposes)** that one script has one hash, and that this hash can act as both its **address payment key** and its **policy ID**:
```

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+    S["one script,<br/>one hash"] -->|as a policy ID| M["`**mint** handler
+    guards making and destroying cards`"]
+    S -->|as an address| P["`**spend** handler
+    guards the funds behind them`"]
+
+    U["the seed UTxO<br/>in your wallet"] -.->|"spent, once"| M
+    M -.->|"same transaction"| P
+```
+
+Because the two handlers share a hash, the spend handler can look at what the transaction mints **under its own policy**. It does not need to be told which policy that is. It can read it from the address it is guarding.
```
rober-m:
```suggestion
Because the two purposes share a hash, the spend purpose can look at what the transaction mints **under its own policy**. It does not need to be told which policy that is. It can read it from the address it is guarding.
```

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+## One card, one contract
```
rober-m, line 146: perfect!

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+  {extractRegion(GiftcardAiken, "giftcard-imports")}
+</CodeBlock>
+
+Then the token name, the two actions, and the two handlers:
+
+<CodeBlock language="aiken" title="validators/giftcard.ak">
+  {extractRegion(GiftcardAiken, "giftcard")}
+</CodeBlock>
+
+There is no datum type in this file, which is answer 1: the token is the state. The two handlers are answer 2, and the body of each one is answer 3.
```
rober-m: change to reference new steps or no steps at all

docs/developers/onboarding/lectures/intermediate/11-multi-validators.md
```diff
@@ -0,0 +1,285 @@
...
+
+Compare the hash with ours:
+
+<CodeBlock>{validatorHash(GiftcardBlueprint, "giftcard.giftcard.mint")}</CodeBlock>
+
+Your seed is still missing, so this is the script with the blank in it. Filling the blank gives a different hash, and that one is the policy ID and the address of a real card.
+
+**Then break it.** Cut the `Create` branch down to its quantity check, dropping the `list.any` call and the `and` around it, and run the tests again. Six still pass, and `create_fails_without_the_seed` is the one that fails. The compiler also warns that `utxo_ref` is now unused. The contract still creates a card, still refuses two, and still releases the funds when a card is burned. It has stopped being a card that can only be created once.
+
+**Now write the check back yourself, without scrolling up.** The rule in words: creating a card is allowed only when the UTxO the contract was built from is among the transaction's inputs.
```
rober-m: I'm not sure if this is necessary, really. Maybe we could tell them to try writing the validator before we show it to them.

Also, I think it'd be helpful to show the transactions (only the tx, no helper functions). They don't need much explanation, but it's better to show them than to have them search the codebase for it.

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+```mermaid
+flowchart LR
+    Seed["a UTxO you own<br/>(the seed)"] -->|"fill the policy's blank<br/>and hash it"| Policy["the beacon's policy ID"]
+    Policy --> Params["the oracle's two blanks<br/>beacon + operator"]
+    Operator["your key hash<br/>(the operator)"] --> Params
+    Params -->|"fill them in<br/>and hash it"| Address["the oracle's address"]
+```
+
+## From idea to contract
```
rober-m: Change to new steps

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+For the app that builds the transaction, this is the unlock you already know, with one addition: one extra output, sent back to the contract's own address, carrying the new datum.
+
+:::note One UTxO, one updater at a time
+A UTxO can only be spent once. So if two updates try to change the same oracle at the same time, one of those transactions fails and has to be built again. That is fine for something a single party publishes. That party does not even have to wait for a block between updates: a transaction's hash is known as soon as it is built, so the next update can spend the continuing output before the previous one confirms, which the handbook calls [transaction chaining](/docs/developers/curriculum/dapps/defi#transaction-chaining). The app in this lecture waits instead, to keep each step visible. Two parties cannot chain past each other, though, and that is why busy contracts store their data in many UTxOs instead of one.
```
rober-m: Great lecture so far! I'd add a link to an explanation of contention. If it doesn't exist, we should create it in the handbook, but I think it exists already.

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+:::
+
+## Which UTxO is the oracle?
+
+A script address is public, and anybody can create a UTxO at one. So anybody can put a UTxO at your oracle's address carrying any rate they like. A contract that reads "the oracle" has no way to tell that one from yours.
```
rober-m:
```suggestion
A script address is public, and anybody can create a UTxO at one. So anybody can put a UTxO at your oracle's address carrying any rate they like. A contract that reads the address as "the oracle" has no way to tell that one from yours.
```

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+The answer is an **NFT**, and you built one in the last lecture. A **[one-shot policy](/docs/developers/onboarding/lectures/intermediate/multi-validators#what-makes-it-an-nft)** is compiled around a **seed** UTxO that the minting transaction has to spend, so the policy can succeed only once, and no second copy of the token can ever exist.
```
rober-m:
```suggestion
In this case, the answer is an **NFT**, and you built one in the last lecture. A **[one-shot policy](/docs/developers/onboarding/lectures/intermediate/multi-validators#what-makes-it-an-nft)** is compiled around a **seed** UTxO that the minting transaction has to spend, so the policy can succeed only once, and when it does, it can only mint one token. So, no second copy of the token can ever exist.
```

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+The gift card used its token as a key: whoever held it could take the funds. The oracle uses the same kind of token as a name. A token used this way is called a **beacon token**, sometimes a state thread token.
```
rober-m:
```suggestion
The gift card used its token as a key: whoever held it could take the funds. The oracle uses the same kind of token as a unique identifier. A token used this way is called a **beacon token** (because it serves as a beacon to find the UTxO) or a **thread token** (because it's threaded between the old and new UTxOs when updating them).
```

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+A token that exists once is not yet an identity. Three rules together make it one:
+
+- The policy mints the beacon once, so there is never a second one to confuse it with.
+- An update has to put the beacon on the output that goes back, so it cannot leave the oracle.
+- Closing the oracle has to burn the beacon, so it cannot outlive the oracle either.
```
rober-m: I wouldn't say these are hard "rules". It depends on the protocol. Becaon/thread tokens don't necessarily have to be NFTs, and in some cases protocol admins may hold them in wallets. 

Maybe I'd remove this paragraph or make it specific to our protocol, not as general rules.

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+    Params -->|"fill them in<br/>and hash it"| Address["the oracle's address"]
+```
+
+## From idea to contract
+
+The same four questions, this time for two validators: the policy that mints the beacon, and the oracle that holds the rate.
+
+**1. What has to be remembered?** The rate, and nothing else, so the datum is a single number. Two more values never change for the whole life of one oracle: the key allowed to publish, which this contract calls the **operator**, and the beacon that names the oracle. By the rule from **[parameters](/docs/developers/onboarding/lectures/intermediate/parameters)**, both are parameters of the oracle. The beacon policy has one parameter of its own, the seed.
+
+**2. What actions are possible?** Four, two in each validator. The policy's are `Mint` and `Burn`, in a `mint` handler. The oracle's are `Update` and `Delete`, in a `spend` handler. Creating the oracle runs only the policy: putting the first output at the oracle's address spends nothing there. As in the **[gift card](/docs/developers/onboarding/lectures/intermediate/multi-validators#how-the-two-handlers-cooperate)**, each validator judges the same transaction on its own.
```
rober-m: The four "actions" are:
- Create Oracle
- Update Oracle
- Read Oracle
- Delete Oracle

Reading Oracle is done by external protocols, so it's not an action done by this protocol, but we have to account for it. That's why we ensure the datum is inline and of the correct type

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+Create `validators/oracle.ak`. The imports first, contract and tests together:
+
+<CodeBlock language="aiken" title="validators/oracle.ak">
+  {extractRegion(OracleAiken, "oracle-imports")}
+</CodeBlock>
+
+Then the types and the validator:
+
+<CodeBlock language="aiken" title="validators/oracle.ak">
+  {extractRegion(OracleAiken, "oracle")}
```
rober-m, line 290: It may be better to remove the `and` and `expect` the shared checks.

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+</CodeBlock>
+
+`Rate` is answer 1, and it is an alias for `Int`: the datum is a bare number with no wrapper around it. `OracleAction` is answer 2. The body is answer 3, and it reads in three parts:
```
rober-m: remvoe reference to old steps

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+1. Find the input being spent, so the contract knows which address to require and what the UTxO came in holding.
+2. The two conditions both actions share sit above the branch: the operator signed, and the UTxO being spent carries the beacon. `holds_beacon` asks the second question.
+3. One branch per action. `Update` requires **exactly one** output going back to that address, reads its datum, confirms the datum is a `Rate`, and requires the beacon to be on it. `Delete` requires the beacon to be burned.
+
+`_datum` is ignored, because the old rate grants nobody anything. Compare the vault, where the datum named the owner and the contract had to read it.
```
rober-m: We ignore the datum's value because we don't use it here. The reader uses it. We just care that it is of the right type

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+Two outputs at the oracle's address would leave the next update with nothing to say which of them is the oracle. The **exactly one** in step 3 is what prevents that.
```
rober-m: No, because one won't have the NFT. We know that we need exactly one, so we can simplify the logic

docs/developers/onboarding/lectures/intermediate/12-modifying-state.md
```diff
@@ -0,0 +1,410 @@
...
+
+1. **Publish rate 100.** One transaction mints the beacon and locks it at the oracle's address with the first rate. Your addresses appear at the top once it goes through. On the explorer, one input is the seed, the mint field shows the beacon with a quantity of 1, and one output sits at the oracle's address holding 5 ADA, the beacon and the datum.
+2. **Refresh**, then **Raise by 50**. Approve it and wait. Refresh again, and the rate reads 150. On the explorer, the input is the old oracle UTxO. One of the outputs is a **new UTxO at the same address**, with a different datum and the same token on it.
+3. **Close.** Approve it and wait. Refresh again, and the app finds nothing. On the explorer, the input is the oracle's last UTxO, the mint field shows the beacon with a quantity of -1, and the 5 ADA is an output at your own address.
+
+Closing is final. The seed is spent, so this beacon can never be minted again. The next oracle you publish starts from a new seed.
+
+## Practise the value rule
```
rober-m, line 366: As explained above, this is not always the case.

rober-m (changes requested):
Checked up until oracle lecture (missing only the last one of intermediate)

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+import OracleBlueprint from "@site/examples/onboarding/lectures/intermediate/oracle/on-chain/aiken/plutus.json";
+
+# Reference inputs & reference scripts
+
+Until now, a transaction could do only one thing with a UTxO: **spend** it. Take it, use it, destroy it. A transaction can also **point at a UTxO** without spending it. The UTxO stays exactly where it is. That single idea has two uses, and they have similar names:
+
+- A **reference script** points at published **code**.
+- A **reference input** points at published **data**.
+
+You need both as soon as another contract wants to use the oracle from the last lecture.
```
rober-m:
```suggestion
You need to reference an input as soon as another contract wants to use the oracle from the last lecture, and we'll use a reference script to make oracle management cheaper and more effective.
```

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Reference scripts
+
+### What it is
+
+A reference script is a compiled contract that has been stored inside a UTxO on the chain. After you store it, a transaction can point at that UTxO instead of carrying its own copy of the contract.
+
+The UTxO that holds the script is an ordinary one at **your own address**. The ADA inside it stays yours. Nothing about the contract changes: same code, same hash, same address, same answers.
```
rober-m:
```suggestion
The UTxO that holds the script is an ordinary one at **any address**. You can use your own address, someone else's, or another contract address, depending on your needs and preferences. Nothing about the contract changes: same code, same hash, same address, same answers.
```

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+
+An unlock that points at it:
```
rober-m: Not sure what you wanted to convey in this transaction, but reference scripts are needed only when the script is being run, and if you're consuming the oracle, the script isn't running. Change the transaction to the same "update oracle" transaction we used in the previous lecture, and add the reference script to it. Also add a short explanation of the tx before the diagram to provide context.

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+
+Skip it for a contract you will run only once or twice. Publishing costs one transaction and locks a small amount of ADA in the UTxO that holds the script, and at very low volume the saving is smaller than that cost.
+
+This is also the only sense in which a Cardano contract is "deployed", a point **[parameters](/docs/developers/onboarding/lectures/intermediate/parameters)** already made.
+
+### What to watch out for
+
+**Pointing at a script still costs something.** Each byte of the referenced script is charged, at a far lower price than carrying the script inside the transaction.
+
+**The UTxO has to stay unspent.** It is an ordinary output that belongs to you, so nothing stops you from spending it. As soon as you do, every transaction that points at it stops working. Publish it, then leave it alone.
```
rober-m:
```suggestion
**The UTxO must stay unspent.** So, if it's an ordinary output that belongs to you, so nothing stops you from spending it. As soon as you do, every transaction that points to it stops working. Publish it somewhere safe, and then leave it alone!
```

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Reference inputs
+
+### What it is
+
+A reference input is a UTxO that a transaction attaches only in order to **read** it. The UTxO stays where it is, and the validator can read its datum.
```
rober-m:
```suggestion
A reference input is a UTxO that a transaction attaches only in order to **read** it. The UTxO stays where it is, and the validator can read its datum, value, address, etc.
```

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+
+Inside the validator, referenced UTxOs arrive in their own field, `reference_inputs`, separate from the ones being spent. You met that field in **[the transaction context](/docs/developers/onboarding/lectures/intermediate/transaction-context)**.
+
+An unlock at the consumer contract, with the oracle attached for reading:
```
rober-m: I don't understand what this transaction tries to do. Please provide context on what the "consumer" does and why it needs the oracle. Based on the code below, say something like, "Let's say the 'consumer' of the oracle will unlock only if the value is greater than zero. In that case, the transaction would look like this"...

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+
+So a transaction has to point at the **current** UTxO. If your app remembers an oracle UTxO from an hour ago and points at it, the transaction is rejected, because that UTxO no longer exists.
+
+**Attaching a UTxO does not make it trustworthy either.** Whoever builds the transaction chooses what to attach, so a validator that reads the first reference input it is handed reads whatever the caller wants it to read. The oracle's beacon is the answer: the consumer searches the reference inputs for that token and reads only the UTxO that carries it.
+
+## Side by side
+
+|  | Reference script | Reference input |
+|---|---|---|
+| so that | transactions stay small | data can be read without being taken |
```
rober-m:
```suggestion
| so that | transactions stay small | data can be read without consuming the UTxO |
```

docs/developers/onboarding/lectures/intermediate/13-reference-inputs-and-scripts.md
```diff
@@ -0,0 +1,288 @@
...
+| it sits at | your own address | the publishing contract's address |
```
rober-m:
```suggestion
| it sits at | any address | the publishing contract's address |
```

rober-m (changes requested):
Reviewed last lecture. I'd say it's ready to merge into our intermediate branch after fixing the pending comments

